PR #1, opened 2017-08-03
DWR-545 Added repository for group_admin_batch.

agorina:


webapp/src/main/java/org/opentestsystem/rdw/admin/model/ImportStatus.java
```diff
@@ -0,0 +1,30 @@
+package org.opentestsystem.rdw.admin.model;
+
+//TODO: it is a copy from ingest common, need to make it shared
```
agorina, line 3: Just wanted to point this out.

webapp/src/test/java/org/opentestsystem/rdw/admin/model/StudentGroupBatchTest.java
```diff
@@ -0,0 +1,61 @@
...
+        assertThat(studentGroupBatch.getId()).isNull();
+        assertThat(studentGroupBatch.getStatus()).isEqualTo(ImportStatus.ACCEPTED);
+        assertThat(studentGroupBatch.getCreator()).isEqualTo("creator");
+        assertThat(studentGroupBatch.getDigest()).isEqualTo("digest");
+        assertThat(studentGroupBatch.getMessage()).isEqualTo("message");
+    }
+
+    @Test
+    public void itShouldNotDefaultACreatedVOrUpdatedalue() {
+        //we want to let DB manage the create/update times.
```
agorina: @malaffoon , is this right?

webapp/src/main/resources/application.sql.yml
```diff
@@ -23,3 +23,17 @@ sql:
...
+  student_group_batch:
+    create: >-
+      INSERT INTO upload_student_group_batch (status, digest, creator, message)
+        VALUES (:status, :digest, :creator, :message)
+
+    findOne: >-
+      SELECT * FROM upload_student_group_batch WHERE id=:id
+
+    findOneByDigest: >-
+      SELECT * FROM upload_student_group_batch WHERE digest=:digest LIMIT 1
```
agorina, line 36: I admit copying it from the import, but I added LIMIT 1 since digest is not unique. Not sure how we want to handle it here?

webapp/src/main/java/org/opentestsystem/rdw/admin/model/ImportStatus.java
chdietz: ok

webapp/src/main/resources/application.sql.yml
chdietz: So digest is not unique in import because of replay, i'm not sure we want that for upload_student_group_batch because we do not want to process the same files twice.  But if there is a code or system error and we need to process the file again we need a way to process it again but nut upload it again?

webapp/src/test/java/org/opentestsystem/rdw/admin/model/StudentGroupBatchTest.java
chdietz: CreatedV ?

malaffoon (approved):

webapp/src/main/java/org/opentestsystem/rdw/admin/model/StudentGroupBatch.java
```diff
@@ -0,0 +1,156 @@
...
+     */
+    public Long getId() {
+        return id;
+    }
+
+    /**
+     * Normal code should not call this, as the data framework will be setting it appropriately.
+     *
+     * @param id system id of import
+     */
```
chdietz: id of StudentGroupBatch ?

webapp/src/main/java/org/opentestsystem/rdw/admin/model/StudentGroupBatch.java
```diff
@@ -0,0 +1,156 @@
...
+    public void setId(final Long id) {
+        this.id = id;
+    }
```
chdietz: When does the data framework set this?

webapp/src/main/resources/application.yml
```diff
@@ -15,7 +15,7 @@ saml:
 
 spring:
   datasource:
-    url: jdbc:mysql://localhost:3306/warehouse?useSSL=false&useLegacyDatetimeCode=false
+    url: jdbc:mysql://localhost:3306/warehouse?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8
     username: root
```
chdietz, line 19: good catch

chdietz (approved):
Not sure about digest and unique still thinking about that.  And the setId? when does the data framework call that?

webapp/src/main/java/org/opentestsystem/rdw/admin/model/StudentGroupBatch.java
agorina: This is copy-and-paste left over, it is not used. Will remove. Thank you!

webapp/src/main/java/org/opentestsystem/rdw/admin/model/StudentGroupBatch.java
agorina: thank you, fixing.

PR #4, opened 2017-08-10
Feature/dwr 737 groups view api

bgsavage:
Added api route for getting groups using a GroupQueryArgumentResolver to resolve query params which downstream maps down to sql query params.

Example routes:
`api/groups?schoolId=1&subjectId=1&schoolYear=2017`
`api/groups?schoolId=1&subjectId=1&subjectId=2&schoolYear=2017`

This is mostly just plumbing, likely will be adding additional data points as needed.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
@@ -1,0 +1,78 @@
+package org.opentestsystem.rdw.admin.model;
+
+public class Group {
+    private String name;
+    private String schoolName;
+    private String subject;
+    private int schoolYear;
+
+    /**
+     * @return The name of the group
+     */
+    public String getName() {
+        return name;
+    }
+
+    /**
+     * @return The name of the school
+     */
+    public String getSchoolName() {
+        return schoolName;
+    }
+
+    /**
+     * @return The subject this group applies to, or NULL if it applies to all subjects
+     */
+    public String getSubject() {
+        return subject;
+    }
+
+    /**
+     * @return The school year for this group
+     */
+    public int getSchoolYear() {
+        return schoolYear;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+        private String name;
+        private String schoolName;
+        private String subject;
+        private int schoolYear;
+
+        public Builder name(final String name) {
+            this.name = name;
+            return this;
+        }
+
+        public Builder schoolName(final String schoolName) {
+            this.schoolName = schoolName;
+            return this;
+        }
+
+        public Builder subject(final String subject) {
+            this.subject = subject;
+            return this;
+        }
+
+        public Builder schoolYear(final int schoolYear) {
+            this.schoolYear = schoolYear;
+            return this;
+        }
+
+        public Group build() {
+            Group group = new Group();
+
+            group.name = name;
+            group.schoolName = schoolName;
+            group.subject = subject;
+            group.schoolYear = schoolYear;
+
+            return group;
+        }
+    }
+}
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupQuery.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupQuery.java
@@ -1,0 +1,65 @@
+package org.opentestsystem.rdw.admin.model;
+
+import java.util.Set;
+
+public class GroupQuery {
+    private int schoolId;
+    private Set<Integer> subjectIds;
+    private int schoolYear;
+
+    /**
+     * @return The school id to query the group by
+     */
+    public int getSchoolId() {
+        return schoolId;
+    }
+
+    /**
+     * @return The subject ids to query the group by
+     */
+    public Set<Integer> getSubjectIds() {
+        return subjectIds;
+    }
+
+    /**
+     * @return The school year to query the group by
+     */
+    public int getSchoolYear() {
+        return schoolYear;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+        private int schoolId;
+        private Set<Integer> subjectIds;
+        private int schoolYear;
+
+        public Builder schoolId(final int schoolId) {
+            this.schoolId = schoolId;
+            return this;
+        }
+
+        public Builder subjectIds(final Set<Integer> subjectIds) {
+            this.subjectIds = subjectIds;
+            return this;
+        }
+
+        public Builder schoolYear(final int schoolyear) {
+            this.schoolYear = schoolyear;
+            return this;
+        }
+
+        public GroupQuery build() {
+            GroupQuery query = new GroupQuery();
+
+            query.schoolId = schoolId;
+            query.subjectIds = subjectIds;
+            query.schoolYear = schoolYear;
+
+            return query;
+        }
+    }
+}
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
@@ -1,0 +1,55 @@
+package org.opentestsystem.rdw.admin.repository.impl;
+
+import com.google.common.collect.ImmutableMap;
+import org.opentestsystem.rdw.admin.common.security.PermissionScope;
+import org.opentestsystem.rdw.admin.model.Group;
+import org.opentestsystem.rdw.admin.model.GroupQuery;
+import org.opentestsystem.rdw.admin.repository.GroupRepository;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+import javax.validation.constraints.NotNull;
+import java.util.List;
+import java.util.Map;
+
+import static org.opentestsystem.rdw.admin.common.jdbc.QueryUtils.getSecurityParameters;
+
+@Repository
+public class JdbcGroupRepository implements GroupRepository {
+    private final NamedParameterJdbcTemplate template;
+
+    @Value("${sql.group.findAllByQuery}")
+    private String findAllQuery;
+
+    @Autowired
+    JdbcGroupRepository(final NamedParameterJdbcTemplate template) {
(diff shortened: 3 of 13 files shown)


webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupQuery.java
```diff
@@ -0,0 +1,65 @@
+package org.opentestsystem.rdw.admin.model;
+
+import java.util.Set;
+
+public class GroupQuery {
```
bgsavage, line 5: Inspired by @malaffoon 's RdwImportQuery object in RDW_Ingest, I find it cleaner to pass around complex types rather than individual method params for a couple of reasons:

1.  It keeps our method parameter lists small
2.  Adding / removing parameters doesn't break the interface / method signatures.  Only need to update the complex type.

webapp/src/main/resources/application.sql.yml
```diff
@@ -9,6 +9,20 @@ sql:
...
+    findAllByQuery: >-
+      select sg.name as group_name,
+        sc.name as school_name,
+        su.code as subject_code,
+        sg.school_year
+      from student_group sg
+        join school sc on sc.id = sg.school_id
+        left join subject su on su.id = sg.subject_id
+      where sg.school_id = :school_id
+        and (sg.subject_id is null or sg.subject_id in (:subject_ids))
```
bgsavage, line 22: subject_id of `null` is inferred to include all subjects, so we always include `null` subject_ids in the query regardless of the subject_ids passed in.  We do this in Reporting as well for permission checking.

webapp/src/test/resources/integration-test-data.sql
```diff
@@ -1,13 +1,16 @@
+insert into import (id, status, content, contentType, digest) values
+  (-1, 1, 1, '', '');
```
bgsavage, line 2: I think I might have been the first one to use this script in RDW_Admin?  A lot of the scripts were broken after recent migrate changes so I updated them and now using this in my new IT

webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupQuery.java
```diff
@@ -0,0 +1,65 @@
...
+    private int schoolId;
```
malaffoon, line 6: Should schoolId and schoolYear be non-primitive types so they can be null, i.e. unspecified in the query?

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public List<Group> findAllByQuery(@NotNull final PermissionScope permissionScope, @NotNull final GroupQuery query) {
+        return template.query(
+                findAllQuery,
+                new MapSqlParameterSource()
+                        .addValues(getSecurityParameters(permissionScope))
+                        .addValues(getQueryParameters(query)),
+                (row, index) -> Group.builder()
+                        .name(row.getString("group_name"))
```
malaffoon, line 40: recommend a static row mapper (like i always do); sure, not reused yet but ...

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
```diff
@@ -0,0 +1,55 @@
...
+                        .schoolYear(row.getInt("school_year"))
+                        .subject(row.getString("subject_code"))
+                        .build()
+        );
+    }
+
+    private static Map<String, Object> getQueryParameters(final GroupQuery query) {
+        return ImmutableMap.<String, Object>builder()
+                .put("school_id", query.getSchoolId())
+                .put("subject_ids", query.getSubjectIds())
```
malaffoon, line 51: If this list is empty will the SQL fail? i.e. does no-subject-ids mean include all or include none?

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultGroupService.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.List;
+
+@Service
+public class DefaultGroupService implements GroupService {
+
+    private final GroupRepository repository;
+
+    @Autowired
+    DefaultGroupService(GroupRepository repository) {
```
malaffoon, line 19: nit: `final`

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultGroupService.java
```diff
@@ -0,0 +1,27 @@
...
+import org.opentestsystem.rdw.admin.repository.GroupRepository;
+import org.opentestsystem.rdw.admin.security.User;
+import org.opentestsystem.rdw.admin.service.GroupService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+
+@Service
+public class DefaultGroupService implements GroupService {
```
malaffoon, line 14: nit: ~~public~~

webapp/src/main/java/org/opentestsystem/rdw/admin/web/GroupController.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.List;
+
+@RestController
+public class GroupController {
+
+    private final GroupService service;
+
+    @Autowired
+    GroupController(GroupService service) {
```
malaffoon, line 20: nit: `final` for params in all methods

webapp/src/main/java/org/opentestsystem/rdw/admin/web/GroupQueryArgumentResolver.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public Object resolveArgument(final MethodParameter parameter,
+                                  final ModelAndViewContainer mavContainer,
+                                  final NativeWebRequest webRequest,
+                                  final WebDataBinderFactory binderFactory) throws Exception {
+        return GroupQuery.builder()
+                .schoolId(parseInt(webRequest.getParameter("schoolId")))
+                .schoolYear(parseInt(webRequest.getParameter("schoolYear")))
+                .subjectIds(Stream
+                        .of(webRequest.getParameterValues("subjectId"))
```
malaffoon, line 34: It would be nice to accept either the subject id or the subject name doing the mapping (via enum helper).

webapp/src/main/resources/application.sql.yml
malaffoon: Ah, this answers a previous question of mine (that i will leave for posterity, as embarrassing as it is).

webapp/src/test/resources/integration-test-data.sql
malaffoon: yeah, likely

webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupQuery.java
```diff
@@ -0,0 +1,65 @@
...
+        private Set<Integer> subjectIds;
+        private int schoolYear;
+
+        public Builder schoolId(final int schoolId) {
+            this.schoolId = schoolId;
+            return this;
+        }
+
+        public Builder subjectIds(final Set<Integer> subjectIds) {
+            this.subjectIds = subjectIds;
```
malaffoon, line 46: should create a new set, copying values

webapp/src/main/java/org/opentestsystem/rdw/admin/web/GroupQueryArgumentResolver.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static java.lang.Integer.parseInt;
+
+
+/**
+ * An argument resolve for the import query object.
+ */
+public class GroupQueryArgumentResolver implements HandlerMethodArgumentResolver {
```
malaffoon, line 19: a unit test for this would be nice

malaffoon (approved):

webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
```diff
@@ -0,0 +1,78 @@
+package org.opentestsystem.rdw.admin.model;
+
+public class Group {
+    private String name;
+    private String schoolName;
+    private String subject;
```
esotrope, line 6: We have```org.opentestsystem.rdw.common.model.Subject``` to represent this too

webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupQuery.java
bgsavage: I think it depends on how robust our api needs to be if we are only building what the UI needs?  If it needs to be robust and we validate, then yes this should be non-primitive...  If not, and we trust the UI won't send bad values, then we are okay here I think.

webapp/src/test/resources/integration-test-data.sql
chdietz: Yes, they were working at one point.

webapp/src/main/resources/application.sql.yml
bgsavage: Still a valid point, if SubjectIds is an empty collection, this SQL fails.  Do we need handle this case?

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
bgsavage: Done, but this and other minor fixes will be updated in my next PR (I've already started working on my next endpoint)

webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
chdietz: That reminds me I wanted to but did not complete remove the rdw-admin-common module, I mentioned this to Brandon.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupQuery.java
chdietz: Sounds good to me, except what Mark mentioned about primitive types.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupQuery.java
chdietz: I would lean in the direction that is more robust, and should we be consistent?

PR #8, opened 2017-08-11
changes for ArchiveService

malaffoon:


webapp/src/main/resources/application.yml
```diff
@@ -36,24 +36,22 @@ spring:
     type: SIMPLE
     cache-names: permissionsByRole
 
-# Needed for access to S3
+# spring auto-configured aws defense
 cloud:
   aws:
-    credentials:
```
malaffoon, line 42: this is confusing to me ... it implies that an S3 instance is being used but it isn't; perhaps just copy&paste stuff

PR #10, opened 2017-08-14
Feature/dwr 737 groups view api

bgsavage:
1.  Deep linking the group filter selections
2.  Added search box
3.  Updated to latest sbac-ui-kit and styles 
4.  Authorizing against "GROUP_READ"
5.  Couple of error case handling
6 . Added the student count column  

image

### DIFF
--- a/webapp/src/main/resources/application.sql.yml
+++ b/webapp/src/main/resources/application.sql.yml
@@ -20,7 +20,8 @@ sql:
       select sg.name as group_name,
         sc.name as school_name,
         sg.subject_id,
-        sg.school_year
+        sg.school_year,
+        (select count(*) from student_group_membership sgm where sgm.student_group_id = sg.id) as student_count
       from student_group sg
         join school sc on sc.id = sg.school_id
       where sg.school_id = :school_id
--- a/webapp/src/main/webapp/package.json
+++ b/webapp/src/main/webapp/package.json
@@ -26,6 +26,7 @@
     "@kourge/ordering": "^1.0.1",
     "@ngx-translate/core": "^6.0.1",
     "@ngx-translate/http-loader": "0.0.3",
+    "@sbac/sbac-ui-kit": "0.0.17",
     "@types/lodash": "^4.14.68",
     "angulartics2": "^2.2.2",
     "core-js": "^2.4.1",
@@ -36,7 +37,6 @@
     "ngx-bootstrap": "^1.6.6",
     "primeng": "^4.1.0",
     "rxjs": "^5.4.2",
-    "sbac-ui-kit": "git+https://github.com/SmarterApp/SBAC-Global-UI-Kit.git#v0.0.10",
     "ts-helpers": "^1.1.1",
     "zone.js": "^0.8.4"
   },
--- a/webapp/src/main/webapp/src/app/app.routes.ts
+++ b/webapp/src/main/webapp/src/app/app.routes.ts
@@ -13,13 +13,15 @@ export const routes: Routes = [
         path: '',
         pathMatch: 'full',
         component: HomeComponent
-      },
-      {
+      }, {
         path: 'groups',
-        pathMatch: 'full',
         resolve: { filterOptions: GroupFilterOptionsResolve },
-        data: { breadcrumb: { translate: 'labels.groups.title' } },
-        component: GroupsComponent
+        children: [ {
+          path: 'filterBy',
+          pathMatch: 'full',
+          data: { breadcrumb: { translate: 'labels.groups.title' } },
+          component: GroupsComponent
+        } ]
       }
     ]
   }
--- a/webapp/src/main/webapp/src/app/app.routes.ts
+++ b/webapp/src/main/webapp/src/app/app.routes.ts
@@ -3,6 +3,7 @@ import { Routes } from "@angular/router";
 import { UserResolve } from "./user/user.resolve";
 import { GroupsComponent } from "./groups/groups.component";
 import { GroupFilterOptionsResolve } from "./groups/group-filter-options.resolve";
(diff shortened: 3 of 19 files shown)


webapp/src/main/resources/application.sql.yml
```diff
@@ -20,7 +20,8 @@ sql:
       select sg.name as group_name,
         sc.name as school_name,
         sg.subject_id,
-        sg.school_year
+        sg.school_year,
+        (select count(*) from student_group_membership sgm where sgm.student_group_id = sg.id) as student_count
```
bgsavage, line 24: Looking for some feedback on how this will perform in aurora.

webapp/src/main/webapp/package.json
```diff
@@ -26,6 +26,7 @@
     "@kourge/ordering": "^1.0.1",
     "@ngx-translate/core": "^6.0.1",
     "@ngx-translate/http-loader": "0.0.3",
+    "@sbac/sbac-ui-kit": "0.0.17",
```
bgsavage, line 29: Now using ci builds for sbac-ui-kit.

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -13,13 +14,17 @@ export const routes: Routes = [
...
+      }, {
         path: 'groups',
-        pathMatch: 'full',
         resolve: { filterOptions: GroupFilterOptionsResolve },
-        data: { breadcrumb: { translate: 'labels.groups.title' } },
-        component: GroupsComponent
+        data: { permissions: [ 'GROUP_READ' ]},
+        canActivate: [ AuthorizeCanActivate ],
+        children: [ {
+          path: 'filterBy',
```
bgsavage, line 23: @wtritch, this is how I solved the issue we were having in Reporting in which only changing the params was causing the parent route to re-resolve it's params.  This was happening because were were using a child path with `''`.  But if I give a name to the child path `'filterBy'` then it works as we expect it to.

webapp/src/main/webapp/src/app/groups/groups.component.ts
```diff
@@ -12,28 +12,54 @@ import { Group } from "./model/group.model";
...
-
     this.query = new GroupQuery(this.filterOptions.subjects);
 
-    this.query.school = this.filterOptions.schools[ 0 ];
-    this.query.schoolYear = this.filterOptions.schoolYears[ 0 ];
-    this.query.subject = this.filterOptions.subjects[ 0 ];
+    if(this.filterOptions.schools.length == 0)
+      return;
+
+    this.route.params.subscribe(p => {
```
bgsavage, line 31: I'm subscribing to the route params so that whenever we change the params the following code will be called.  Otherwise on back / forward button clicks the params were not updating (since I'm using a named child route)

webapp/src/main/webapp/src/app/groups/groups.component.spec.ts
```diff
@@ -16,26 +18,37 @@ describe('GroupsComponent', () => {
       return Observable.of([])
     }
   };
-  let filterOptions = new GroupFilterOptions();
+  let filterOptions: GroupFilterOptions;
+  let mockRouteSnapshot;
+  let paramsObserver: Observer<{}>;
```
esotrope: What is ```Observer<{}>``` is that the same as ```Observer<Object>```?

webapp/src/main/webapp/src/app/groups/groups.component.ts
```diff
@@ -12,28 +12,54 @@ import { Group } from "./model/group.model";
...
     this.service
       .getGroups(this.query)
       .subscribe(groups => {
         this.groups = groups;
+        this.filterGroups();
       })
   }
+
+  filterGroups() {
+    this.filteredGroups = this.groups.filter( x => x.name.toUpperCase().indexOf(this.searchTerm.toUpperCase()) >= 0)
```
bgsavage, line 63: Note:  Searches will be case insensitive.

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -30,16 +30,19 @@
...
     },
     "groups": {
-      "title": "Groups",
+      "title": "Student Group Administration",
       "select-school": "School",
       "select-school-year": "School Year",
       "select-subject": "Subject",
-      "no-results": "No groups found",
+      "no-results": "No groups found matching the given filter criteria.",
+      "no-schools": "Your user account does not have access to view the groups of any schools.",
```
bgsavage, line 38: Looking for feedback on this message.  Technically it would be a bad setup for you to have a `GROUPS_READ` permission and have no schools scoped to that permission.

webapp/src/main/webapp/src/primeng-overrides.less
```diff
@@ -0,0 +1,195 @@
+//
```
bgsavage, line 1: Brought this file in from Reporting

webapp/src/main/webapp/src/styles.less
```diff
@@ -3,109 +3,30 @@
 // Extending SBAC UI Kit
```
bgsavage, line 3: Updated to what we have in Reporting

webapp/src/main/webapp/src/app/groups/groups.component.spec.ts
bgsavage: Yeah maybe I should put `any` here actually.

webapp/src/main/webapp/src/assets/i18n/en.json
esotrope: Maybe we can add the "please see your administrator if you feel your access is incorrect" message?
If it's a page view we might just want to throw up a 404 page no?

PR #13, opened 2017-08-16
UI & API for deleting a group

bgsavage:
image

 - Requires GROUP_WRITE permission scoped to that school in order to delete.
 - Soft delete implemented as a restful PUT
 - New unit / integration tests still coming.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
@@ -9,7 +9,7 @@ public class Group {
     private Subject subject;
     private int schoolYear;
     private int studentCount;
-    private boolean isDeleted;
+    private boolean deleted;
 
     /**
      * @return The group id
@@ -56,8 +56,8 @@ public class Group {
     /**
      * @return true if this group is deleted
      */
-    public boolean getIsDeleted() {
-        return isDeleted;
+    public boolean isDeleted() {
+        return deleted;
     }
 
     public static Builder builder() {
@@ -71,7 +71,7 @@ public class Group {
         private Subject subject;
         private int schoolYear;
         private int studentCount;
-        private boolean isDeleted;
+        private boolean deleted;
 
         public Builder id(final int id) {
             this.id = id;
@@ -103,8 +103,8 @@ public class Group {
             return this;
         }
 
-        public Builder isDeleted(final boolean isDeleted) {
-            this.isDeleted = isDeleted;
+        public Builder deleted(final boolean isDeleted) {
+            this.deleted = isDeleted;
             return this;
         }
 
@@ -117,7 +117,7 @@ public class Group {
             group.subject = subject;
             group.schoolYear = schoolYear;
             group.studentCount = studentCount;
-            group.isDeleted = isDeleted;
+            group.deleted = deleted;
 
             return group;
         }
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
@@ -3,11 +3,20 @@ package org.opentestsystem.rdw.admin.model;
 import org.opentestsystem.rdw.common.model.Subject;
 
 public class Group {
+    private int id;
     private String name;
     private String schoolName;
     private Subject subject;
     private int schoolYear;
     private int studentCount;
+    private boolean isDeleted;
+
+    /**
+     * @return The group id
+     */
+    public int getId() {
+        return id;
+    }
 
     /**
      * @return The name of the group
@@ -44,16 +53,30 @@ public class Group {
         return studentCount;
     }
 
+    /**
+     * @return true if this group is deleted
+     */
+    public boolean getIsDeleted() {
+        return isDeleted;
+    }
+
     public static Builder builder() {
         return new Builder();
     }
 
     public static class Builder {
+        private int id;
         private String name;
         private String schoolName;
         private Subject subject;
         private int schoolYear;
         private int studentCount;
+        private boolean isDeleted;
+
+        public Builder id(final int id) {
+            this.id = id;
+            return this;
+        }
 
         public Builder name(final String name) {
             this.name = name;
@@ -80,14 +103,21 @@ public class Group {
             return this;
         }
 
+        public Builder isDeleted(final boolean isDeleted) {
+            this.isDeleted = isDeleted;
+            return this;
+        }
+
         public Group build() {
             final Group group = new Group();
 
+            group.id = id;
             group.name = name;
             group.schoolName = schoolName;
             group.subject = subject;
             group.schoolYear = schoolYear;
             group.studentCount = studentCount;
+            group.isDeleted = isDeleted;
 
             return group;
         }
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupSet.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupSet.java
@@ -1,0 +1,46 @@
+package org.opentestsystem.rdw.admin.model;
+
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+public class GroupSet {
+    private boolean writeable;
+    private Set<Group> groups;
+
+    public boolean getWriteable() {
+        return writeable;
+    }
+
+    public Set<Group> getGroups() {
+        return groups;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+        private boolean writeable;
+        private Set<Group> groups;
+
+        public Builder writeable(final boolean writeable) {
+            this.writeable = writeable;
+            return this;
+        }
+
+        public Builder groups(final Set<Group> groups) {
+            this.groups = ImmutableSet.copyOf(groups);
+            return this;
+        }
+
+        public GroupSet build() {
+            final GroupSet groupSet = new GroupSet();
+
+            groupSet.writeable = writeable;
+            groupSet.groups = groups;
+
+            return groupSet;
+        }
+    }
+}
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupSet.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupSet.java
@@ -8,10 +8,16 @@ public class GroupSet {
     private boolean writeable;
     private Set<Group> groups;
 
-    public boolean getWriteable() {
+    /**
+     * @return true if the current user has the GROUP_WRITE permission for this set of groups
+     */
+    public boolean isWriteable() {
         return writeable;
     }
 
+    /**
+     * @return set of groups
+     */
     public Set<Group> getGroups() {
         return groups;
     }
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupRepository.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupRepository.java
@@ -21,4 +21,11 @@ public interface GroupRepository {
      * @return all the distinct school years for groups this user has access to.
      */
     List<Integer> findAllDistinctSchoolYears(@NotNull PermissionScope permissionScope);
+
+    /**
+     * Updates this group
+     * @param permissionScope   The user's permission scope
+     * @param group             The group being updated
+     */
+    void update(@NotNull PermissionScope permissionScope, @NotNull Group group);
 }
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
@@ -61,7 +61,7 @@ public class JdbcGroupRepository implements GroupRepository {
         );
     }
 
-    // TODO: This implementation only updates the isDeleted field.
+    // TODO: This implementation only updates the deleted field.
     // TODO: Other fields need to be validated before they can be updated.
     @Override
     public void update(@NotNull final PermissionScope permissionScope, @NotNull final Group group) {
@@ -84,6 +84,7 @@ public class JdbcGroupRepository implements GroupRepository {
                 .schoolYear(row.getInt("school_year"))
                 .subject(parseSubject(row.getString("subject_id")))
                 .studentCount(row.getInt("student_count"))
+                .deleted(row.getBoolean("deleted"))
                 .build();
     }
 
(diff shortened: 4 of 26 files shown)


webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupSet.java
```diff
@@ -0,0 +1,46 @@
+package org.opentestsystem.rdw.admin.model;
+
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+public class GroupSet {
```
bgsavage, line 7: Had to create a wrapper which has the `writeable` property which indicates if the user is authorized to edit (and delete) this group.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupSet.java
```diff
@@ -0,0 +1,46 @@
...
+    private boolean writeable;
+    private Set<Group> groups;
+
+    public boolean getWriteable() {
```
bgsavage: I need javadocs here.

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
```diff
@@ -57,8 +61,24 @@
...
+    public void update(@NotNull final PermissionScope permissionScope, @NotNull final Group group) {
+        final int result = template.update(
+                update,
+                new MapSqlParameterSource()
+                        .addValues(getSecurityParameters(permissionScope))
+                        .addValues(getGroupParameters(group))
+        );
+
+        if(result == 0)
+            throw new NoSuchElementException();
```
bgsavage, line 76: If nothing was updated, throwing an exception.  At this point, the UI shouldn't be showing the "Delete" link for groups the user cannot delete, so this shouldn't happen, but just in case.

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
```diff
@@ -79,6 +99,13 @@ private static Group mapResultToGroup(final ResultSet row) throws SQLException {
                 .build();
     }
 
+    private static Map<String, Object> getGroupParameters(final Group group) {
+        return ImmutableMap.<String, Object>builder()
+                .put("id", group.getId())
+                .put("deleted", group.getIsDeleted())
```
bgsavage: I'm only implementing the the `deleted` field at this point.  When we come later (in phase 2?) to implement Edit, we can just include the other group fields once they've been validated.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultGroupService.java
```diff
@@ -28,8 +30,15 @@
...
-    public Set<Group> getByQuery(final User user, final GroupQuery query) {
-        return ImmutableSet.copyOf(repository.findAllByQuery(user.getPermissionsById().get("GROUP_READ").getScope(), query));
+    public GroupSet getByQuery(final User user, final GroupQuery query) {
+        final Permission writePermission = user.getPermissionsById().get("GROUP_WRITE");
+        final Permission readPermission = user.getPermissionsById().get("GROUP_READ");
+
+        return GroupSet
+                .builder()
+                .groups(ImmutableSet.copyOf(repository.findAllByQuery(readPermission.getScope(), query))                )
+                .writeable(writePermission != null && schoolService.existsByPermissionScope(writePermission.getScope(), query.getSchoolId()))
```
bgsavage, line 40: It's possible for a user to have GROUP_READ for a group but not have GROUP_WRITE, so we need to make this check agains the given schoolId.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
```diff
@@ -3,11 +3,20 @@
 import org.opentestsystem.rdw.common.model.Subject;
 
 public class Group {
+    private int id;
     private String name;
     private String schoolName;
     private Subject subject;
     private int schoolYear;
     private int studentCount;
+    private boolean isDeleted;
```
malaffoon: convention in java would have this as `deleted` with the getter as `isDeleted`

webapp/src/main/resources/application.sql.yml
```diff
@@ -14,21 +14,33 @@ sql:
...
+        and 1=:statewide or sc.district_id in (:district_ids) or sg.school_id in (:school_ids)
 
     findAllByQuery: >-
-      select sg.name as group_name,
+      select sg.id,
+        sg.name as group_name,
         sc.name as school_name,
         sg.subject_id,
         sg.school_year,
+        sg.active,
```
bgsavage: Hm, I don't need active anymore.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
```diff
@@ -44,16 +53,30 @@ public int getStudentCount() {
...
     }
 
     public static class Builder {
+        private int id;
         private String name;
         private String schoolName;
         private Subject subject;
         private int schoolYear;
         private int studentCount;
+        private boolean isDeleted;
```
malaffoon: not sure there is a convention for builders but i would make this `deleted` with the method `deleted` as well

webapp/src/main/resources/application.sql.yml
```diff
@@ -14,21 +14,33 @@ sql:
...
       where sg.school_id = :school_id
+        and sg.deleted = 0
         and (sg.subject_id is null or sg.subject_id in (:subject_ids))
         and sg.school_year = :school_year
         and (1=:statewide or sc.district_id in (:district_ids) or sg.school_id in (:school_ids))
 
+    update: >-
+      update student_group sg
+        join school s on sg.school_id = s.id
+      set sg.deleted = :deleted
```
bgsavage, line 39: Just re-iterating here, I'm only updating `deleted` field right now and should be updated once we implement edit groups.

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupRepository.java
```diff
@@ -21,4 +21,11 @@
      * @return all the distinct school years for groups this user has access to.
      */
     List<Integer> findAllDistinctSchoolYears(@NotNull PermissionScope permissionScope);
+
+    /**
+     * Updates this group
```
malaffoon: suggest putting a comment about "just updating deleted flag" here

webapp/src/main/webapp/src/app/groups/delete-group.modal.ts
```diff
@@ -0,0 +1,29 @@
...
+  constructor(private bsModalRef: BsModalRef, private service: GroupService) {
+  }
+
+  cancel() {
+    this.bsModalRef.hide();
+  }
+
+  save() {
+    this.group.isDeleted = true;
+    this.service.update(this.group).subscribe(res => {
```
bgsavage, line 23: UI implementation of "deleting a group".

webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -53,7 +53,13 @@
           </ng-template>
         </p-column>
         <p-column field="studentCount" header="{{'labels.groups.cols.student-count' | translate}}" [sortable]="true"></p-column>
-
+        <p-column *ngIf="isAuthorizedToWrite" field="isActive">
```
bgsavage: This field needs to be updated to isDeleted

webapp/src/main/java/org/opentestsystem/rdw/admin/service/SchoolService.java
```diff
@@ -19,4 +20,12 @@
...
     Set<Organization> getSchools(@NotNull User user);
 
+    /**
+     * Checks if a school exists by the given permission scope.
+     *
+     * @param permissionScope  the permission scope to check
+     * @param schoolId          the school id to check
+     * @return true if the school exists and is part of the permission scope.
+     */
+    boolean existsByPermissionScope(@NotNull PermissionScope permissionScope, int schoolId);
```
malaffoon: Do we really need `ByPermissionScope` in the method name? Seems pretty obvious from the required parameter.

malaffoon (approved):

webapp/src/main/webapp/src/app/groups/groups.module.ts
```diff
@@ -6,14 +6,22 @@ import { GroupService } from "./groups.service";
...
+import { DeleteGroupModalComponent } from "./delete-group.modal";
+
 @NgModule({
   declarations: [
-    GroupsComponent
+    GroupsComponent,
+    DeleteGroupModalComponent
+  ],
+  entryComponents: [
+    DeleteGroupModalComponent
```
bgsavage, line 18: This threw me off for a while when my modal wasn't working.  So for modal components we need to explicitly call it out in the `entryComponents` because it does not have a selector reference, is not part of a route definition, and is not part of bootstrap.  So the only way angular will know to add it to the componentFactory is if we call it out here in the `entryComponents`

webapp/src/main/webapp/src/app/groups/groups.service.ts
```diff
@@ -51,15 +61,31 @@ export class GroupService {
...
+  private mapGroupToApi(uiModel: Group): any {
+    let apiModel: any = {};
+
+    apiModel.id = uiModel.id;
+    apiModel.schoolYear = uiModel.schoolYear;
+    apiModel.name = uiModel.name;
+    apiModel.schoolName = uiModel.schoolName;
+    apiModel.subject = uiModel.subject;
+    apiModel.studentCount = uiModel.studentCount;
+    apiModel.isDeleted = uiModel.isDeleted;
```
bgsavage: I am sending the entire group object back to the API though, in anticipation that at some point the API will be saving all of this data.

webapp/src/main/webapp/src/app/shared/data/data.service.ts
```diff
@@ -1,16 +1,13 @@
...
 import "rxjs/add/operator/map";
 import "rxjs/add/operator/catch";
-import { CachingDataService } from "../cachingData.service";
 
-// TODO: Break out methods from DataService so only a public generic Get.
-// TODO: Other methods such as getGroups belong in their own service such as GroupService
 @Injectable()
 export class DataService {
 
-  constructor(private http: Http, private staticDataService: CachingDataService) {
```
bgsavage, line 13: unrelated: not used

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
```diff
@@ -57,8 +61,24 @@
         );
     }
 
+    // TODO: This implementation only updates the isDeleted field.
+    // TODO: Other fields need to be validated before they can be updated.
+    @Override
+    public void update(@NotNull final PermissionScope permissionScope, @NotNull final Group group) {
```
jtreuting: Right now it seems the UI is setting the deleted flag and doing a normal update.  The soft-delete feels like an implementation detail where the front-end should just say it wants to delete it and the service deletes it.  It feels to me like these update methods should be delete and the API should use a DELETE verb, and then we just happen to be implementing that delete as a soft-delete by updating a specific flag in the DB.

bgsavage: @malaffoon and I discussed this.  Since the UI will eventually have to become aware of the soft delete by allowing an "undelete", then a PUT would seem to work best since you can use that same PUT to do the undelete, instead of having to implement a `groups/id/undelete` as a separate api route.

bgsavage:
I'm closing this PR as it needs to be reworked since we should only ever be looking at the GROUP_WRITE permission and never the GROUP_READ permission in order to view groups.  This will simplify the API and UI we will no longer have to worry about the case of a user viewing a group that they don't have access to delete.

webapp/src/main/webapp/src/app/groups/delete-group.modal.ts
esotrope: I would think here that "update" is an implementation detail of soft delete and we would actually want to call this service method delete no?

webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
```diff
@@ -3,11 +3,20 @@
 import org.opentestsystem.rdw.common.model.Subject;
 
 public class Group {
+    private int id;
```
esotrope, line 6: If this is an entity ID, generally we have been using type long for good measure

webapp/src/main/resources/application.sql.yml
```diff
@@ -40,6 +51,14 @@ sql:
       from school
       where (1=:statewide or district_id in (:district_ids) or id in (:school_ids))
 
+    exists: >-
+      select exists(
+        select 1
+        from school
+        where id = :id
+          and (1=:statewide or district_id in (:district_ids) or id in (:school_ids))
```
esotrope, line 59: ```limit 1``` will short circuit this select I think

webapp/src/main/webapp/src/app/groups/delete-group.modal.html
```diff
@@ -0,0 +1,23 @@
...
+  </div>
+  <div *ngIf="unableToDelete" class="alert alert-danger">
+    {{ 'labels.groups.delete.error' | translate }}
+  </div>
+  <div class="modal-footer">
+    <span>
+      <button type="button" class="btn btn-default" (click)="cancel()">
+        {{ (unableToDelete? 'prompt.close' : 'labels.groups.delete.button-cancel') | translate }}
+      </button>
+      <button type="button" *ngIf="!unableToDelete" class="btn btn-primary" (click)="save()">
```
esotrope, line 18: I think we might want to use btn-danger here but I'll delegate to the UI team to be sure

webapp/src/main/resources/application.sql.yml
bgsavage: Do we need to short circuit if I'm searching on a primary key?

esotrope: Good point. You are just searching to match 1 indexed ID so it will make no difference 👍

webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
bgsavage: I can make this a long, though you know I really long for ints... :)

esotrope: 🌽 🧀 🌽 🙄

webapp/src/main/webapp/src/app/groups/delete-group.modal.html
bgsavage: I like how this looks, thanks.

@aaronsleeper thoughts??

image

aaronsleeper: looks good

PR #14, opened 2017-08-16
Delete Group and Requiring GROUP_WRITE permission

bgsavage:
Sorry for the confusion.

This PR now only uses the "GROUP_WRITE" permission, ignoring "GROUP_READ".  This simplifies the implementation in that we no longer need to worry about groups a user can view but not delete.  Now checking the GROUP_WRITE permission before activating the home component.  If a user comes to the admin portal without the GROUP_WRITE permission, then they will get an error message.


webapp/src/main/java/org/opentestsystem/rdw/admin/model/GroupSet.java
```diff
@@ -0,0 +1,52 @@
+package org.opentestsystem.rdw.admin.model;
+
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+public class GroupSet {
```
bgsavage: ignore: it will be deleted shortly.  I this would have been removed when I did my revert commit.

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -17,8 +19,6 @@ export const routes: Routes = [
       }, {
         path: 'groups',
         resolve: { filterOptions: GroupFilterOptionsResolve },
-        data: { permissions: [ 'GROUP_READ' ]},
-        canActivate: [ AuthorizeCanActivate ],
```
bgsavage, line 21: moved to parent, and changed to GROUP_WRITE

webapp/src/main/java/org/opentestsystem/rdw/admin/model/Group.java
```diff
@@ -80,14 +103,21 @@ public Builder studentCount(final int studentCount) {
             return this;
         }
 
+        public Builder deleted(final boolean isDeleted) {
```
esotrope: nit: isDeleted -> deleted

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
```diff
@@ -79,6 +100,13 @@ private static Group mapResultToGroup(final ResultSet row) throws SQLException {
...
+    private static Map<String, Object> getGroupParameters(final Group group) {
+        return ImmutableMap.<String, Object>builder()
+                .put("id", group.getId())
+                .put("deleted", group.isDeleted())
+                .build();
+    }
+
     private static Subject parseSubject(final String id) {
         try {
             return id == null
```
esotrope, line 112: Just want to make sure - null is interpreted as both subjects by the UI right?

webapp/src/main/java/org/opentestsystem/rdw/admin/web/GroupController.java
```diff
@@ -27,9 +31,14 @@
         return service.getByQuery(user, query);
     }
 
+    @PutMapping("/api/groups")
```
esotrope, line 34: Do we want to use delete semantics all the way up until the repo implementation and let update/soft delete be the impl or do we want the client to know that it is soft deleting?
I'm not sure what philosophy we have here.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/GroupService.java
```diff
@@ -8,9 +8,27 @@
 
 import java.util.Set;
 
-@PreAuthorize("hasAuthority('PERM_GROUP_READ')")
+@PreAuthorize("hasAuthority('PERM_GROUP_WRITE')")
```
esotrope, line 11: Aside from service lock-down with these annotations I think we can more restrictively update the spring ```*SecurityConfig.java``` file to only permit access to ```/*``` to users with permission ```"PERM_GROUP_WRITE"```

webapp/src/main/java/org/opentestsystem/rdw/admin/web/GroupController.java
bgsavage: The idea here is that at some point we would like the UI to be able to see deleted groups and have the ability to "undelete" them, so the soft deletion implementation would have to leak all the way to the front end, at which point this would just become another update.  This is why I'm doing this as an update.  This also sets us up for when we do edit groups.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/GroupService.java
bgsavage: Could this admin portal have other functions besides just group at some point?

Also I'd still like some basic endpoints such as user and translate to flow through so we can at least render an error page to the user should they not have access.

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/impl/JdbcGroupRepository.java
bgsavage: Yep, we have a subject pipe which does just that.

webapp/src/main/java/org/opentestsystem/rdw/admin/web/GroupController.java
wtritch: Hmmm... I would consider soft-deleting an implementation detail that the API should be unaware of, my vote would be for DELETE semantics all the way through.

PR #16, opened 2017-08-17
operational settings

malaffoon:


webapp/build.gradle
```diff
@@ -4,6 +4,10 @@ plugins {
 
 apply plugin: 'org.springframework.boot'
 
+createDockerfile {
+    ext.java_opts = ["-Xmx512m"]
```
malaffoon, line 8: I'll set container memory limit to 800m to go with this.

PR #17, opened 2017-08-17
Feature/dwr 767 ui ux

bgsavage:
The Student Group Administration is now the home page.
The file upload page now a new upload dropzone, table, and progress indicator.

image

image


webapp/src/main/webapp/src/app/admin/admin.component.html
```diff
@@ -1,101 +0,0 @@
-<div class="home-view">
```
bgsavage, line 1: this file was moved but is pretty much different now anyway.

webapp/src/main/webapp/src/app/app.module.ts
```diff
@@ -12,19 +11,18 @@ import { routes } from "./app.routes";
...
     AppComponent,
-    BreadcrumbsComponent,
-    HomeComponent
+    BreadcrumbsComponent
   ],
   imports: [
     BrowserModule,
     CommonModule,
-    AdminModule,
+    GroupImportModule,
```
bgsavage, line 25: Renamed the module as `AdminModule` is all encompassing so thought to target the feature area in case we have other admin functions in this app unrelated to groups.

webapp/src/main/webapp/src/app/app.routes.ts
```diff
@@ -1,29 +1,38 @@
...
+    resolve: { user: UserResolve, filterOptions: GroupFilterOptionsResolve },
+    data: { permissions: [ 'GROUP_WRITE' ] },
     canActivate: [ AuthorizeCanActivate ],
     children: [
       {
         path: '',
-        pathMatch: 'full',
-        component: HomeComponent
+        pathMatch: 'prefix',
+        redirectTo: 'groups'
```
bgsavage, line 18: This is different then reporting.  Unfortunately in angular we are not able to have route params on the root route, which I'm using to provide deep linking of the basic filters.  So I'm having to redirect to another route.  In the long run this is probably better anyway in case we should have more functionality other than just groups in the admin app, this allows us to create a home component rather easily to link to the groups page without upsetting all the routing of the app.

webapp/src/main/webapp/src/app/breadcrumbs/breadcrumbs.component.html
```diff
@@ -1,13 +1,7 @@
 <!-- Breadcrumb -->
-<div [hidden]="breadcrumbs.length == 0">
+<div [hidden]="breadcrumbs.length <= 1">
```
bgsavage, line 2: If we do actually have a home component later, we would want this to be 0 again, and add back in the home route link I removed below.

webapp/src/main/webapp/src/app/groups/import/group-import.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  ngOnInit() {
+    this.uploader = new FileUploader({ url: URL });
+    this.uploader.setOptions({ autoUpload: true});
+
+    this.uploader.onCompleteItem = (item: any, response: any, status: any, headers: any) => {
+      console.log(item);
+      if (!isNullOrUndefined(response)) {
+
+        let importResult = this.studentGroupService.mapStudentGroupBatchFromApi(JSON.parse(response));
+        importResult.fileName = item.file.name;
```
bgsavage: Grabbing the filename from the event.  This will need to be persisted for when we want to see this on the history page.

webapp/src/main/webapp/src/app/groups/import/group-import.component.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  ngOnInit() {
+    this.uploader = new FileUploader({ url: URL });
+    this.uploader.setOptions({ autoUpload: true});
+
+    this.uploader.onCompleteItem = (item: any, response: any, status: any, headers: any) => {
+      if (!isNullOrUndefined(response)) {
+
+        let importResult = this.studentGroupService.mapImportResultFromApi(JSON.parse(response));
+        importResult.fileName = item.file.name;
```
bgsavage, line 33: Grabbing the filename from the uploader.  This will need to be persisted in order for the history to show this.

PR #18, opened 2017-08-17
Feature/validate upload

wtritch:
This PR performs front-end validation during the initial upload before shipping the CSV off to the group-processor for migration/import into the data mart.

In order to validate the CSV as a stream, I'm not requiring that the CSV be sorted by School, Group Name, and School Year.  The user will see validation failure messages if the CSV is out of order.

Validations I'm performing are:
1. Confirm that the user has GROUP_WRITE access to the school
2. Confirm that the CSV has all required headers
3. Confirm that each CSV record contains the School, Group Name, and School Year required to uniquely identify a group.
4. Confirm that a Group's subject is set to either Null, All, Math, or ELA.  And, once set, is not changed to a different subject by another record.

Now that we are no longer chunking, I don't require any schema changes for this PR.  The upload represents a single Batch, is written to S3 during upload, and has an associate entry in the upload_student_group_batch table created prior to sending the "Go" RabbitMQ message to the group-processor.


webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvSubject.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.admin.model;
+
+/**
+ * This class represents the Group subjects allowed during CSV Group uploads.
+ */
+public enum CsvSubject {
```
wtritch, line 6: Unfortunately, since "All" is a viable subject for the CSV export, I couldn't re-use the org.opentestsystem.rdw.common.model.Subject enum

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/SchoolRepository.java
```diff
@@ -17,7 +17,7 @@
      * @param permissionScope the scope of access
      * @return all of the schools accessible through the given permission scope
      */
-    List<Organization> findAll(@NotNull PermissionScope permissionScope);
+    List<School> findAll(@NotNull PermissionScope permissionScope);
```
wtritch: Since I needed the school.natural_id column, I had to extend Organization in order to retrieve it from the DB.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
```diff
@@ -0,0 +1,61 @@
...
+    final static String HeaderSSID = "student_ssid";
+
+    final static Set<String> RequiredHeaders = ImmutableSet.of(
+            HeaderSchool,
+            HeaderSubject,
+            HeaderGroup,
+            HeaderUser,
+            HeaderYear,
+            HeaderSSID
+    );
```
wtritch: Since these are used across services/tests, I decided to centralize them here.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/StudentGroupBatchService.java
```diff
@@ -10,7 +10,7 @@
 /**
  * Service responsible for managing {@link StudentGroupBatch}s
  */
-//TODO: adjust permissions appropriately
+@PreAuthorize("hasAuthority('PERM_GROUP_WRITE')")
```
wtritch, line 13: Changed this service to require GROUP_WRITE permissions and moved the annotation to the class level.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,200 @@
...
+                    .row(0)
+                    .message("Missing headers " + missing + ", found " + headers)
+                    .build());
+        }
+        return failures;
+    }
+
+    @Override
+    public List<CsvValidationFailure> validateRecords(final Iterator<CSVRecord> recordIterator,
+                                                      final PermissionScope permissionScope) {
```
wtritch, line 74: Here I'm accepting a streaming Iterator of CSVRecords.  By requiring that the CSV be ordered, I can validate on a stream and don't need any all-knowing memory constructs that could potentially blow the heap.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
```diff
@@ -110,27 +109,75 @@ public StudentGroupBatch upload(final User user, final MultipartFile file) {
...
+                             final String digest,
+                             final User user) throws IOException {
+        try (final InputStream fileStream = file.getInputStream()) {
+            final String location = new LocationStrategy.GroupUploadContentLocationStrategy().location(digest);
+
+            //Check if file already exists in ArchiveService
+            if (archiveService.exists(location)) {
+                logger.debug("Upload already exists in S3: {}", file.getOriginalFilename());
+                return;
+            }
```
wtritch, line 133: If the file already exists with the same hash, I don't re-upload it to S3.  I could see a user wanting to "reset" their group permissions by re-uploading the same file, but there's no reason to re-send the same bits to S3.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
bgsavage: These feels a bit "smurfy" to me.  Should they be in their own constant class?

webapp/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationServiceTest.java
```diff
@@ -0,0 +1,233 @@
...
+    private static Map<String, Integer> createHeaderMap() {
+        final List<String> orderedHeaders = newArrayList(RequiredHeaders);
+        final Map<String, Integer> headerMap = new HashMap<>();
+        for (int i=0; i < orderedHeaders.size(); i++) {
+            headerMap.put(orderedHeaders.get(i), i);
+        }
+        return headerMap;
+    }
+
+    private static class RecordBuilder {
```
wtritch: This test harness was a pita to set up.  It annoys me that the Apache CSVRecord is a "final" class with a package-local constructor.  It means that I need to actually create a CSV in order to test.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,200 @@
...
+
+                validateSchool(validationFailures, availableSchools, completedSchools, school, record.getRecordNumber());
+            }
+
+            //Validate the School, Group, and Year
+            final String schoolGroupYear = school + group + String.valueOf(schoolYear);
+            if (currentSchoolGroupYear == null || !currentSchoolGroupYear.equalsIgnoreCase(schoolGroupYear)) {
+                if (currentSchoolGroupYear != null) completedSchoolGroupYears.add(currentSchoolGroupYear);
+                currentSchoolGroupYearSubject = null;
+                currentSchoolGroupYear = schoolGroupYear;
```
bgsavage, line 113: Hm, I sense a pattern of very similar code here, any chance to extract a method for reuse?

webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
wtritch: I was resistant as well.  But apparently the industry like constants on Interfaces rather than implementations.  I could break them into a separate constant class, I suppose...  CsvGroupHeaders.java?

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
wtritch: I fought with it, but although they look similar, every line is different.  Sometimes I setting one thing null, sometimes 2, sometimes I'm clearing a map... I couldn't find common abstract ground.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,200 @@
...
+                    .row(recordNumber)
+                    .message("Record is out-of-order: sort CSV by school, group, year")
+                    .build());
+        }
+    }
+
+    @Override
+    public String toMessage(final List<CsvValidationFailure> failures) {
+        final StringBuilder messageBuilder = new StringBuilder();
+        int maxMessages = MaxFailureMessages;
```
bgsavage: The naming of `maxMessages` is confusing, maybe should be a `messageCount` ?

wtritch: Good call, changed

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,200 @@
...
+
+        if (schoolYear == null) {
+            validationFailures.add(CsvValidationFailure.builder()
+                    .row(recordNumber)
+                    .message("Record does not contain a school year")
+                    .build());
+        }
+    }
+
+    private void validateSchool(final List<CsvValidationFailure> validationFailures,
```
bgsavage, line 193: Hm, might be cleaner to follow SRP here and break each of these validation methods into their own classes that all implement some `Validator` interface.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/School.java
```diff
@@ -0,0 +1,23 @@
+package org.opentestsystem.rdw.admin.model;
+
+import org.opentestsystem.rdw.admin.common.model.Organization;
+
+/**
+ * This model represents a school, containing the id, name, and natural id.
+ */
+public class School extends Organization {
+
+    private String naturalId;
```
malaffoon: I'm pretty darn sure that every Organization has a naturalId ... that should be put there.

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/SchoolRepository.java
malaffoon: See note above.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
```diff
@@ -65,43 +66,41 @@ public DefaultStudentGroupBatchService(@NotNull final StudentGroupBatchRepositor
...
+                return repository.create(studentGroupBatchBuilder.build());
             }
 
-            final String location = new LocationStrategy.GroupUploadContentLocationStrategy().location(digest);
-            archiveService.writeResource(location, fileStream, properties);
+            logger.debug("Validated request: {}", file.getOriginalFilename());
+            archiveFile(file, digest, user);
 
             studentGroupBatchBuilder
                     .digest(digest)
```
bgsavage: would be nice to add the digest before line 87 if we always need it.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
```diff
@@ -65,43 +66,41 @@ public DefaultStudentGroupBatchService(@NotNull final StudentGroupBatchRepositor
...
-            if (sgbImport != null) {
-                logger.info("Ignoring file name {} with existing digest {}", file.getName(), digest);
+            final String digest = Hex.encodeHexString(messageDigest.digest()).toUpperCase();
+            if (!validationFailures.isEmpty()) {
+                studentGroupBatchBuilder
+                        .digest(digest)
+                        .message(validationService.toMessage(validationFailures))
+                        .status(ImportStatus.BAD_DATA);
+                logger.debug("Request failed validation: {}", file.getOriginalFilename());
+                return repository.create(studentGroupBatchBuilder.build());
```
bgsavage: This if block maybe really should only set status and message, and do the create outside since we always call the repo.create too.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
malaffoon: Constants in the Interface makes sense iff the constants are used as part of the interface. 

If they are an implementation detail then i suggest a separate class. If it's just cause of tests i'd bury them in the implementation and make the tests reach in to get them.

bgsavage: Sounds good to me.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,200 @@
...
+
+        while (recordIterator.hasNext()) {
+            final CSVRecord record = recordIterator.next();
+            final String school = record.get(HeaderSchool);
+            final String group = record.get(HeaderGroup);
+            final Integer schoolYear = getAsInt(record, HeaderYear);
+            final CsvSubject subject = CsvSubject.valueOfCaseInsensitive(record.get(HeaderSubject));
+
+            if (isBlank(school) && isBlank(group) && schoolYear == null) {
+                //Not a viable record, skip
```
malaffoon: Do we want to at least count records like this?

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    @Override
+    public String toMessage(final List<CsvValidationFailure> failures) {
+        final StringBuilder messageBuilder = new StringBuilder();
+        int maxMessages = MaxFailureMessages;
+        for (final CsvValidationFailure failure : failures) {
+            messageBuilder.append(String.format(ValidationFormat, failure.getRow(), failure.getMessage()));
+            if (maxMessages-- < 0) {
```
malaffoon: This whole thing is a bit confusing; it seems like a subList could be used instead. Meh.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
malaffoon: Yeah, the whole ignore-the-same-content thing only makes sense if the data isn't changed any other way.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/School.java
wtritch: ick, that's going to spiderweb.  I'll look into it.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
wtritch: I meant to call this out.  Does this make sense?  Or should I just pull this check out?  I think we're set to *not* ignore empty lines, but *should* we ignore empty lines?

malaffoon: I'd suggest ignoring empty lines; i've seen samples (not for this exact payload but...) where empty lines are used to separate things. Then, of course, any line with delimiters and no data could be considered bad data.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
wtritch: That's much cleaner, changed.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvSubject.java
malaffoon: Hmm, i'd lean more towards a parseSubject helper method that returns Subject and knows that `All` == `` == null.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
wtritch: Hm... except that it's stateful validation, with validators modifying state and order being important.  That feels like it might reduce readability with hidden dependencies between validators.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
wtritch: The digest is calculated by pulling bits through the DigestInputStream on line 82, we can't get it until after we've read the whole uploaded file.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
wtritch: Then we'd have to shove everything below into an else block (or additional if) since we don't want to modify the StudentGroupBatchBuilder or attempt to archive the file if the upload failed validation.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
wtritch: If something else is downloading and re-uploading a file to the same location (based upon hashed content) then we probably have a larger issue =P

wtritch: @malaffoon : question, if an upload to S3 fails part-way through (say, network failure) the entire file upload is dropped, right?  We don't end up with a corrupted file in S3?

malaffoon: In the one small test i did that is exactly what happened ... no file appeared in S3.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvSubject.java
wtritch: Except that All != null.  If no subject is specified then (All/Null) is assumed, and can be set to Math/ELA by the next record.  If All is specified, and the next record sets the subject to Math/ELA, that's a validation failure.  So we do actually have 4 "acceptable" subject values: null, All, Math, and ELA.

malaffoon: Oh. Yeah. 

So, if the subject is null for all records for a group, are we assuming it is like All was specified somewhere? Or is it an error?

webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
wtritch: well shoot, was only used by the implementation and tests.  Moved to the implementation.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvSubject.java
wtritch: I believe that we're assuming that it's as if All is specified somewhere, but that's the responsibility of the group-processor @chdietz

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
wtritch: That makes me happy.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
bgsavage: Yep, I'd lean something closer to....

```
if(validationFailures.IsEmpty()){
    setSuccessState(builder);
}
else {
    setErrorState(builder); 
}
```

Maybe just personal preference, but I think it's easier to follow methods that have a single return (unless it's a short circuit at the top) rather than try and hunt down all the places a method could return. 

Anyways, just a thought... :)

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
wtritch: That sounds more reasonable than hand-jamming ignore cases.  Changed to ignore empty lines.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
wtritch: I restructured a bit with validation success/failure if/else and submission to rabbit at the bottom.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/School.java
wtritch: Wasn't too bad.  Done.

PR #20, opened 2017-08-18
Minor cleanup of error handling and messages provided to user

wtritch:
Fixing some minor issues found during initial testing/development.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
@@ -63,7 +63,7 @@ public class DefaultCsvValidationService implements CsvValidationService {
             missing.removeAll(headers);
             failures.add(CsvValidationFailure.builder()
                     .row(0)
-                    .message("Missing headers " + missing + ", found " + headers)
+                    .message("Missing headers " + missing)
                     .build());
         }
         return failures;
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -63,7 +63,7 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
             missing.removeAll(headers);
             failures.add(CsvValidationFailure.builder()
                     .row(0)
-                    .message("Missing headers " + missing + ", found " + headers)
+                    .message("Missing headers " + missing)
```
wtritch, line 66: There's no reason to show the user their existing headers.  They have the file, they know what the headers are.  We just need to tell them which headers we need that they are missing.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
```diff
@@ -170,14 +170,21 @@ private void archiveFile(final MultipartFile file,
...
                     .row(0)
                     .message("Error reading file [" + ex.getMessage() + "]")
                     .build());
+        } catch (final IllegalArgumentException ex) {
+            logger.warn("Problem parsing uploaded file", ex);
+            validationFailures.clear();
+            validationFailures.add(CsvValidationFailure.builder()
+                    .row(0)
+                    .message("File does not appear to be a valid CSV with a header row")
+                    .build());
```
wtritch, line 187: When the CSVParser encounters bad data (say, from a .docx or .mp4 file) it throws an IllegalArgumentException containing data from the source file.  To protect against having a lot of unknown characters in the response message, I'm going to log the error and return a generic "probably not a CSV with a header row" response message.

malaffoon (approved):

PR #23, opened 2017-08-18
Feature/ui tweaks

bgsavage:
- Now toggling the View Errors button -> Hide Errors when clicked
image

- Now defaulting to first param in view group when param is now a filter option.


wtritch:
Should we only allow 1 row to be expanded at a time for consistency?

bgsavage:
I thought about this.  We have this limitation in Reporting bc of loading multiple IRIS iframes.  I'm not sure I should restrict an admin user from comparing error messages across rows because of a limitation in Reporting.

PR #24, opened 2017-08-18
Add validation check on the number of students per group.  Maximum is

wtritch:
200 students per group.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
@@ -20,6 +20,7 @@ import java.util.stream.Collectors;
 
 import static com.google.common.collect.Sets.newHashSet;
 import static org.apache.commons.lang.StringUtils.isBlank;
+import static org.apache.commons.lang.StringUtils.isNotBlank;
 
 /**
  * Default implementation of a CsvValidationService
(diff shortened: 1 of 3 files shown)


webapp/src/test/java/org/opentestsystem/rdw/admin/CsvGenerator.java
```diff
@@ -0,0 +1,116 @@
...
+import static org.opentestsystem.rdw.admin.service.impl.DefaultCsvValidationService.HeaderSubject;
+import static org.opentestsystem.rdw.admin.service.impl.DefaultCsvValidationService.HeaderUser;
+import static org.opentestsystem.rdw.admin.service.impl.DefaultCsvValidationService.HeaderYear;
+import static org.opentestsystem.rdw.admin.service.impl.DefaultCsvValidationService.RequiredHeaders;
+
+/**
+ * This testing utility class can be used to generate arbitrarily-large CSV sample files.
+ * Replace the school ids list with valid values from your local database.
+ */
+public class CsvGenerator {
```
wtritch, line 30: I decided to upload this little generator utility in case anyone else would like to generate test CSVs.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -46,6 +47,7 @@
     private static final String ValidationFormat = "Row: %1$d Failure: %2$s\n";
     private static final String MoreFailuresFormat = "With %1$d additional validation failures\n";
     private static final int MaxFailureMessages = 100;
+    static final int MaxStudentsPerGroup = 200;
```
chdietz, line 50: May not be an immediate need but I am wondering if this should be configurable?
Probably not for phase 1 anyway because this is the agreed hard limit.

webapp/src/test/java/org/opentestsystem/rdw/admin/CsvGenerator.java
chdietz: nice

chdietz (approved):

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
wtritch: It could be, but we should either discuss how to ensure the configuration value is shared between the admin and group-processor projects.  I suppose we could always just add comments saying that the two properties must be changed in lock-step.  But I think you're right, that may be a phase 2 issue.  The client may want this UX to completely change in phase 2.

PR #25, opened 2017-08-18
Feature/ui tweaks

bgsavage:
- Now translating import status
- Now using the filterable dropdown for schools
- Now warning users from navigating away while upload is in progress


webapp/src/main/webapp/src/app/groups/import/group-import.deactivate.ts
```diff
@@ -0,0 +1,19 @@
+import { CanDeactivate } from '@angular/router';
+import { GroupImportComponent } from "./group-import.component";
+import { Injectable } from "@angular/core";
+import { TranslateService } from "@ngx-translate/core";
+
+@Injectable()
+export class GroupImportDeactivateGuard implements CanDeactivate<GroupImportComponent> {
```
bgsavage, line 7: This only guards against navigation within the angular app.

webapp/src/main/webapp/src/app/groups/import/group-import.component.ts
```diff
@@ -35,6 +36,16 @@ export class GroupImportComponent implements OnInit {
 
     this.uploader.onCompleteAll = () => {
       this.uploader.clearQueue();
+    };
+
+    window.onbeforeunload = this.confirmNavigation.bind(this);
```
bgsavage, line 41: This will guard against navigation outside the angular app (i.e. tab close or url change)

webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -105,6 +107,16 @@
...
+    },
+    "status": {
+      "UNKNOWN_SCHOOL": "Unknown School",
+      "UNKNOWN_ASMT": "Unknown Assessment",
+      "UNAUTHORIZED": "Unauthorized",
+      "BAD_DATA": "Bad Data",
+      "BAD_FORMAT": "Bad Format",
+      "INVALID": "Invalid",
+      "ACCEPTED": "Accepted",
+      "PROCESSED": "Processed"
```
bgsavage, line 119: @wtritch @chdietz @rcordeau  I realize the status enum was just a copy / paste from Ingest.  Should you add / remove statuses please update their translate values here.

PR #26, opened 2017-08-22
adds link to home page reporting webapp (DWR-824)

esotrope:
Adds ability to set where the SBAC logo links to.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/AppSettings.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/AppSettings.java
@@ -1,8 +1,16 @@
 package org.opentestsystem.rdw.admin.model;
 
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Component
 public class AppSettings {
 
+    @Value("${app.analytics.trackingId}")
     private String analyticsTrackingId;
+
+    @Value("${app.reporting.webapp.url}")
+    private String reportingWebappURL;
 
     private AppSettings() {
     }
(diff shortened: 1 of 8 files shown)


jtreuting (approved):
The thinking here is that we can have it link back to the Reporting application if we want this to be part of the flow when they click Manage Student Groups instead of opening a new window/tab from the reporting app when they click that?

esotrope:
@jtreuting - yes. Also sensibly defaults to just go home on the admin app if we choose not to go with sending the user back to reporting.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/AppSettings.java
```diff
@@ -1,9 +1,17 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Component
 public class AppSettings {
 
+    @Value("${app.analytics.trackingId}")
     private String analyticsTrackingId;
 
+    @Value("${app.reporting.webapp.url}")
```
malaffoon: I think the property should be `app.reporting-webapp-url`. The only reason to have it `app.reporting.webapp.url` would be if there were other properties that needed to be set like `app.reporting.fubar.shoesize` and that isn't really the case here: this is the reportingWebappUrl, a single property.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/AppSettings.java
```diff
@@ -1,9 +1,17 @@
...
+    private String reportingWebappURL;
```
malaffoon: I think this should be `reportingWebappUrl`. Whether i personally like it or not, the convention is to fully camel-case even for acronyms and initializations.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/AppSettings.java
esotrope: k I'll adjust it

webapp/src/main/resources/application.yml
```diff
@@ -89,6 +89,10 @@ permissionservice:
   endpoint: ${rdw.permission.service}
 
 app:
+  reporting:
+    webapp:
+      url: /
```
malaffoon: Should this be http://localhost/ or some such?  I guess i don't know what works for local devs but, for my own edification, this is expected to be a fully URL, correct?  e.g. https://rdw-prod-reporting/?

malaffoon: Oh, just saw your comment to JT about navigating to home in the current app by default.

esotrope: Good point.
I realized actually now that you wrote this I need to change the variable name to be ```home-url``` or something. The idea here is that clicking the logo will by default take you to the home page of the admin app OR if configured to, take you back to the reporting webapp. And in the default case it's best to use "/" so we can take advantage of the browser's host fill-in

malaffoon: Ah, it's even more complicated than i knew. I will quietly back out of the room and let the UX folks weigh in.

PR #27, opened 2017-08-23
Migrate group deletions to the data-mart

wtritch:
This PR creates an import record on group deletion, sets the deleted group's update_import_id to the new import record, then marks the import as ready.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupImportRepository.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupImportRepository.java
@@ -6,8 +6,8 @@ package org.opentestsystem.rdw.admin.repository;
 public interface GroupImportRepository {
 
     /**
-     * Create a Import record to represent a warehouse
-     * to data-mart Group migration.
+     * Create an Import record to associate with new group data;
+     * the record is created with status ACCEPTED
      *
      * @param digest    The import digest (max 32-character string)
      * @param batch     The import batch (max 250-character string)
@@ -16,9 +16,9 @@ public interface GroupImportRepository {
     long create(String digest, String batch);
 
     /**
-     * Trigger the given Import.
+     * Update the given Import record to the PROCESSED status.
      *
      * @param importId The import id
      */
-    void trigger(long importId);
+    void setProcessed(long importId);
 }
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupImportRepository.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupImportRepository.java
@@ -1,0 +1,24 @@
+package org.opentestsystem.rdw.admin.repository;
+
+/**
+ * This repository is responsible for working with imports.
+ */
+public interface GroupImportRepository {
+
+    /**
+     * Create a Import record to represent a warehouse
+     * to data-mart Group migration.
+     *
+     * @param digest    The import digest (max 32-character string)
+     * @param batch     The import batch (max 250-character string)
+     * @return The created import id
+     */
+    long create(String digest, String batch);
+
+    /**
+     * Trigger the given Import.
+     *
+     * @param importId The import id
+     */
+    void trigger(long importId);
+}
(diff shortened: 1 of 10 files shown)


webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultGroupService.java
```diff
@@ -48,7 +56,11 @@ public GroupFilterOptions getFilterOptions(final User user) {
     }
 
     @Override
+    @Transactional
     public void delete(final User user, final long id) {
-        repository.delete(user.getPermissionsById().get("GROUP_WRITE").getScope(), id);
+        final String digest = HexUtils.toHexString(DigestUtils.md5(user.getEmail() + id));
+        final long importId = importRepository.create(digest, "group delete");
+        repository.delete(user.getPermissionsById().get("GROUP_WRITE").getScope(), id, importId);
+        importRepository.trigger(importId);
```
wtritch: This is the big change.  By marking the service method as transactional I don't need to handle fallback cases where the group doesn't exist or the user doesn't have access.

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupImportRepository.java
```diff
@@ -0,0 +1,24 @@
...
+     * @return The created import id
+     */
+    long create(String digest, String batch);
+
+    /**
+     * Trigger the given Import.
+     *
+     * @param importId The import id
+     */
+    void trigger(long importId);
```
malaffoon: nit: the verb `trigger` seems loaded. I'm assuming all this does is mark the import record as PROCESSED? Why not a method name that reflects that?

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupImportRepository.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * This repository is responsible for working with imports.
+ */
+public interface GroupImportRepository {
+
+    /**
+     * Create a Import record to represent a warehouse
+     * to data-mart Group migration.
+     *
```
malaffoon: Perhaps additional verbiage that says something like, "Create an Import record to associate with new group data; the record is created with status ACCEPTED"

webapp/src/main/resources/application.sql.yml
```diff
@@ -67,3 +68,13 @@ sql:
 
     sqlFindByCreator: >-
       SELECT * FROM upload_student_group_batch WHERE creator=:creator
+
+  import:
+    createQuery: >-
+      INSERT INTO import (status, content, contentType, digest, batch) VALUES
+        (0, 5, 'admin group modification', :digest, :batch)
```
malaffoon, line 75: I'm not thrilled with the ongoing abuse of the `contentType` column. I don't really have a better suggestion just bringing it up as a discussion topic.

malaffoon (approved):

webapp/src/main/resources/application.sql.yml
agorina: I am of the same opinion. But I have to admit that due to the lack of a better field and time pressure I am doing the same with the legacy data load.

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupImportRepository.java
wtritch: works for me, done

webapp/src/main/java/org/opentestsystem/rdw/admin/repository/GroupImportRepository.java
wtritch: works for me, done

webapp/src/main/resources/application.sql.yml
wtritch: I wasn't sure what to actually shove in there, the digest, or batch columns since this isn't a file-based import...  I feel like I'm breaking assumptions, but I don't have a better short-term alternative.

PR #31, opened 2017-08-24
Feature/dwr 711

rcordeau:


webapp/build.gradle
```diff
@@ -39,6 +39,8 @@ dependencies {
 
     compile 'org.apache.commons:commons-csv:+'
 
+    compile 'com.google.code.gson:gson'
```
wtritch, line 42: I'd prefer using the existing jackson ObjectMapper for json serialization/deserialization.

PR #32, opened 2017-08-25
DWR-929 Validate CSV header order on upload

wtritch:
With the CSV-to-DB load script in place, the CSV header order is now required to be known.  This adds CSV header order validation during the CSV upload process.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
@@ -6,7 +6,7 @@ import org.opentestsystem.rdw.admin.model.CsvValidationFailure;
 
 import java.util.Iterator;
 import java.util.List;
-import java.util.Set;
+import java.util.Map;
 
 /**
  * Implementations of this interface are responsible for validating
@@ -18,10 +18,10 @@ public interface CsvValidationService {
      * Validate the given CSV's headers and return the collection
      * of validation failures.
      *
-     * @param csvHeaders CSV headers
+     * @param csvHeaders The map of CSV header to index
      * @return The list of validation failures, empty if valid
      */
-    List<CsvValidationFailure> validateHeaders(Set<String> csvHeaders);
+    List<CsvValidationFailure> validateHeaders(Map<String, Integer> csvHeaders);
 
     /**
      * Validate the given CSV records and return the collection
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
@@ -6,7 +6,6 @@ import org.opentestsystem.rdw.admin.model.CsvValidationFailure;
 
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 
 /**
  * Implementations of this interface are responsible for validating
@@ -18,10 +17,10 @@ public interface CsvValidationService {
      * Validate the given CSV's headers and return the collection
      * of validation failures.
      *
-     * @param csvHeaders The map of CSV header to index
+     * @param csvHeaders The ordered list of CSV headers
      * @return The list of validation failures, empty if valid
      */
-    List<CsvValidationFailure> validateHeaders(Map<String, Integer> csvHeaders);
+    List<CsvValidationFailure> validateHeaders(List<String> csvHeaders);
 
     /**
      * Validate the given CSV records and return the collection
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
@@ -1,6 +1,7 @@
 package org.opentestsystem.rdw.admin.service.impl;
 
-import com.google.common.collect.ImmutableSet;
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
 import org.apache.commons.csv.CSVRecord;
 import org.opentestsystem.rdw.admin.common.model.Organization;
 import org.opentestsystem.rdw.admin.common.security.PermissionScope;
@@ -15,10 +16,10 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import static com.google.common.collect.Sets.newHashSet;
 import static org.apache.commons.lang.StringUtils.isBlank;
 import static org.apache.commons.lang.StringUtils.isNotBlank;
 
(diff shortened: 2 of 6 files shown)


webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -28,20 +29,20 @@
...
+    public final static List<String> RequiredHeaders = ImmutableList.of(
             HeaderGroup,
-            HeaderUser,
+            HeaderSchool,
             HeaderYear,
-            HeaderSSID
+            HeaderSubject,
+            HeaderSSID,
+            HeaderUser
     );
```
wtritch, line 45: This is the order required the the CSV load SQL statement.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -57,16 +58,20 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
...
+            final String header = RequiredHeaders.get(i);
+            if (!csvHeaders.containsKey(header) || csvHeaders.get(header) != i) {
+                failures.add(CsvValidationFailure.builder()
+                        .row(0)
+                        .message("Invalid headers. Headers must be in order: [" +
+                                Joiner.on(",").join(RequiredHeaders) +
+                                "]")
+                        .build());
+                break;
+            }
```
wtritch, line 72: The CSVParser gives us a Map<String, Integer> containing the header names and their index.  This just validates that the headers are in the expected order.

webapp/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationServiceTest.java
```diff
@@ -56,22 +56,41 @@ public void setup() {
...
+            headerMap.put(misOrderedHeaders.get(i), i);
+        }
 
-        final CsvValidationFailure failure = service.validateHeaders(missingHeaders).get(0);
+        final CsvValidationFailure failure = service.validateHeaders(headerMap).get(0);
 
         assertThat(failure.getRow()).isEqualTo(0);
-        assertThat(failure.getMessage()).contains(HeaderSchool);
+        assertThat(failure.getMessage()).contains(Joiner.on(",").join(RequiredHeaders));
     }
```
wtritch: Here's the validating test.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
esotrope: Is there a ```parser.getHeaderList()``` ? Seems like it would be more straight forward to compare the order in which the headers appear with the order the required headers are in.
```if (!csvHeaders.get(i).equals(requiredHeaders.get(i))) { /* wrong order */ }```

File: webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java [hunk 4]
```diff
@@ -18,10 +18,10 @@
      * Validate the given CSV's headers and return the collection
      * of validation failures.
      *
-     * @param csvHeaders CSV headers
+     * @param csvHeaders The map of CSV header to index
      * @return The list of validation failures, empty if valid
      */
-    List<CsvValidationFailure> validateHeaders(Set<String> csvHeaders);
+    List<CsvValidationFailure> validateHeaders(Map<String, Integer> csvHeaders);
```
malaffoon: Uhm, why not pass a List<String> in? A list is ordered.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -28,20 +29,20 @@
...
 
-    public final static String HeaderSchool = "school_natural_id";
-    public final static String HeaderSubject = "subject_code";
     public final static String HeaderGroup = "group_name";
-    public final static String HeaderUser = "group_user_login";
+    public final static String HeaderSchool = "school_natural_id";
     public final static String HeaderYear = "school_year";
+    public final static String HeaderSubject = "subject_code";
     public final static String HeaderSSID = "student_ssid";
+    public final static String HeaderUser = "group_user_login";
```
esotrope, line 36: do we plan on aliasing these so that they make more sense to a user? or do we like that they map directly to db columns for troubleshooting or something?

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
wtritch: I wish.  But no, it was the first thing I looked for.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
wtritch: I suppose I could, and do the map-to-list conversion in the client.  The problem is that the CSVParser only gives me a map, so I have to do the map-to-list conversion somewhere.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
esotrope: Well that's just wacksauce

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
wtritch: They map directly to DB columns based upon index.  I suppose we could name them whatever we like, but why increase complexity for CSVs that shouldn't be hand-jammed in "real life" ?

esotrope: Oh ok. Based on the rule page i wrote i was under the impression that people would be hand jamming these to some degree but i guess we have some level of automation setup for them?

wtritch: ideally they'll be exporting them from a school system of some kind.  At least that's my understanding.

chdietz: yeah, we really do not know. we know districts export similar data to CALPADS so I would assume whoever does that could write an export to produce the CSV we have defined.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/CsvValidationService.java
wtritch: changed

esotrope (approved):

PR #34, opened 2017-08-26
Change service interface to accept a list of headers rather than a map

wtritch:
This PR fixes an out of bounds exception being thrown during CSV validation


webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
```diff
@@ -173,10 +175,9 @@ private void archiveFile(final MultipartFile file,
                 return validationFailures;
             }
 
-            final List<String> orderedHeaders = new ArrayList<>(parser.getHeaderMap().size());
-            for (final Map.Entry<String, Integer> entry : parser.getHeaderMap().entrySet()) {
-                orderedHeaders.add(entry.getValue(), entry.getKey());
-            }
```
wtritch, line 179: Oops.  I had mistakenly assumed that initializing the list with a capacity allowed me to insert up to that capacity without an index out-of-bounds exception.
Switched to insert-and-sort method below.

PR #37, opened 2017-08-31
version 1.0.0

malaffoon:
The CI process is such that develop should have the version number set to what we're working towards, in this case 1.0.0.

I've already bumped common and schema.

Note that i am not resetting the build number cause it really should've been 1.0.0 for a while.

### DIFF
--- a/webapp/src/main/webapp/package.json
+++ b/webapp/src/main/webapp/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "client",
+  "name": "admin",
   "version": "1.0.0",
   "license": "MIT",
   "angular-cli": {},
--- a/webapp/src/main/webapp/package.json
+++ b/webapp/src/main/webapp/package.json
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "0.0.0",
+  "version": "1.0.0",
   "license": "MIT",
   "angular-cli": {},
   "scripts": {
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "0.0.0",
+  "version": "1.0.0",
```
malaffoon: Is this right? Is the client version used/shown anywhere?

chdietz: Also wondering if there is a package-lock.json for RDW_Admin

rcordeau: Yes we need to update the package-lock.json file too.

chdietz: I do not see where this is displayed on the client.

rcordeau: I don't think it is displayed, but we should probably be consistent.

bgsavage: do we want to give an actual `name` at this point as well?

rcordeau: How about testApp1?

bgsavage: I prefer `robTestApp14`.  Shows effort.

PR #2, opened 2017-08-04
Feature/dwr 545

agorina:

PR #39, opened 2017-08-31
fixes modal event subscription issues

esotrope:


webapp/src/main/webapp/src/app/groups/groups.component.ts
```diff
@@ -1,85 +1,109 @@
...
+    this.filteredGroups = this.groups
+      .filter(x => x.name.toUpperCase().indexOf(this.searchTerm.toUpperCase()) >= 0);
   }
 
-  openSetActiveModal(group: Group) {
-    this.bsModalRef = this.modalService.show(DeleteGroupModalComponent);
-    this.bsModalRef.content.group = group;
-    this.modalService.onHide.subscribe(x => {
-      this.updateResults();
-    });
```
esotrope, line 83: This was the root of the issue and was problematic for multiple reasons.
1) we need to clean up the subscription otherwise it will add a new subscription and callback every opening of the same modal.
2) the modal hiding doesn't exactly equate to a deletion having been made, so the group list was updating even when the modal was cancelled
3) we know which group is deleted successfully so we don't need to ping the backend for the reports we can just remove the one we deleted from the list

webapp/src/main/webapp/src/app/groups/groups.component.ts
```diff
@@ -1,85 +1,109 @@
...
               private modalService: BsModalService) {
   }
 
+  get groups(): Group[] {
+    return this._groups;
+  }
+
+  set groups(groups: Group[]) {
+    this._groups = groups;
+    this.updateFilteredGroups();
```
esotrope, line 37: every time the groups update the filtered selection needs to update so it seemed logical to put this here.

webapp/src/main/webapp/src/app/groups/groups.component.ts
bgsavage: For 2 and 3, my thoughts were just to always refresh the data in case something changed while that modal was up.  For #1, I think we should just unsubscribe on the OnDestroy maybe?

esotrope: Every time subscribe was called on the modalService it was accumulating. I think under the hood when we subscribe to the modal service it holds the event emitter at the service level and doesn't just delegate to the modal instance emitters.

For #2/3 i think if we lookup recent status of the groups when they happen to delete them, they might think they deleted more than just what they selected no?

PR #43, opened 2017-09-08
Changed app title to 'Manage Student Groups' and removed unused trans…

bgsavage:
…lations.

Also updated welcome message for consistency with reporting app


webapp/src/main/webapp/src/assets/i18n/en.json
```diff
@@ -1,7 +1,7 @@
 {
   "welcome": "Welcome {{firstName}} {{lastName}}",
   "application": {
-    "title": "Student Group Administration"
+    "title": "Manage Student Groups"
```
bgsavage, line 4: This is the real change, everything else is unrelated while I'm already in this project.

esotrope (approved):

PR #44, opened 2017-09-09
Now showing the school's natural id in the school dropdown

bgsavage:
image

### DIFF
--- a/webapp/src/main/webapp/src/app/groups/groups.component.ts
+++ b/webapp/src/main/webapp/src/app/groups/groups.component.ts
@@ -46,7 +46,7 @@ export class GroupsComponent implements OnInit, OnDestroy {
     }
 
     this.schoolDropdownOptions = this.filterOptions.schools.map(school => <any>{
-      label: school.name,
+      label: `${school.name} (${school.naturalId})`,
       value: school
     });
 
--- a/webapp/src/main/webapp/src/app/groups/groups.component.ts
+++ b/webapp/src/main/webapp/src/app/groups/groups.component.ts
@@ -47,7 +47,9 @@ export class GroupsComponent implements OnInit, OnDestroy {
 
     this.schoolDropdownOptions = this.filterOptions.schools.map(school => <any>{
       label: `${school.name} (${school.naturalId})`,
-      value: school
+      value: school,
+      name: school.name,
+      naturalId: school.naturalId
     });
 
     this.route.params.subscribe((params: any) => {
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/src/app/groups/model/school.model.ts
```diff
@@ -1,4 +1,5 @@
 export class School {
   id: number = 0;
   name: string = "";
```
esotrope, line 3: why do these have defaults?

webapp/src/main/webapp/src/app/groups/groups.component.ts
```diff
@@ -46,7 +46,7 @@ export class GroupsComponent implements OnInit, OnDestroy {
     }
 
     this.schoolDropdownOptions = this.filterOptions.schools.map(school => <any>{
-      label: school.name,
+      label: `${school.name} (${school.naturalId})`,
```
esotrope: Wonder if we should de-emphasize the natural ID in the parens by making it gray. Or maybe it's best to keep it black since it's searchable. Not sure.

webapp/src/main/webapp/src/app/groups/model/school.model.ts
bgsavage: They probably aren't needed.

webapp/src/main/webapp/src/app/groups/groups.component.ts
bgsavage: I can definitely de-emphasize and keep the natural id searchable using an ngTemplate.

@aaronsleeper @esotrope  what do you think?

image

esotrope: The word wrap looks pretty awful but real school IDs probably aren't going to be that long i guess. Other than that I think it looks a little nicer.

aaronsleeper: will users know the NaturalID?
if yes, and it's necessary, then I think it should be on a second line and have .small applied to de-emphasize it.

bgsavage: Can do.  Should I still keep it in parens?

aaronsleeper: nah. not necessary

bgsavage: image

aaronsleeper:
perf. make it one or two shades lighter .gray-dark should do it i think

PR #45, opened 2017-09-09
DWR-831 Check for extra headers past required headers to prevent erro…

chdietz:
The original error reported by DWR-831 mentions an error writing to S3 and I expect this is a system error.  We should think about errors message in this category.

Turns out there is a problem if the valid headers exist but there is an extra column, or at least an extra header at the end.  When deployed on AWS group processor will fail loading from S3 with an extra header, this does not appear to be the case in local environment.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
@@ -71,6 +71,14 @@ public class DefaultCsvValidationService implements CsvValidationService {
                 break;
             }
         }
+        if ( csvHeaders.size() > RequiredHeaders.size() ) {
+            failures.add(CsvValidationFailure.builder()
+                    .row(0)
+                    .message("Invalid headers. Additional headers exist after required headers: [" +
+                            Joiner.on(",").join(RequiredHeaders) +
+                            "]")
+                    .build());
+        }
         return failures;
     }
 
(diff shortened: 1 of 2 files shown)


webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -71,6 +71,14 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
                 break;
             }
         }
+        if ( csvHeaders.size() > RequiredHeaders.size() ) {
```
malaffoon, line 74: I'd suggest that you put something before the loop checking the headers. It looks like the existing code will throw if RequiredHeaders > csvHeaders. And now we're adding a condition that there can't be extra headers. So i would suggest, before line 63:
```
    if (csvHeaders.size() != RequiredHeaders.size()) {
        failures.add(...);
        return failures;
    }
```

webapp/src/test/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationServiceTest.java
```diff
@@ -74,6 +74,18 @@ public void itShouldNotValidateMissingHeaders() {
...
+        final List<String> additionalHeaders = newArrayList(RequiredHeaders);
+        additionalHeaders.add("extra_header_at_end");
+
+        final CsvValidationFailure failure = service.validateHeaders(additionalHeaders).get(0);
+
+        assertThat(failure.getRow()).isEqualTo(0);
+        assertThat(failure.getMessage()).contains("Additional headers");
+        assertThat(failure.getMessage()).contains(Joiner.on(",").join(RequiredHeaders));
+    }
+
```
malaffoon, line 87: Can you add a test where the found headers are the same as the required headers except for the missing last one?
```
    final List<String> headers = newArrayList(RequiredHeaders);
    headers.remove(headers.size()-1);
    ...
```

PR #46, opened 2017-09-12
Now adding success validation results on each successful school with …

bgsavage:
…a message indicating how many groups were accepted for that school.

Updated UI to now show success / error messages in the grid as panel

image
image

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
@@ -4,10 +4,11 @@ package org.opentestsystem.rdw.admin.model;
  * This model represents a CSV Validation failure containing the failing row
  * and failure message.
  */
-public class CsvValidationFailure {
+public class CsvValidationResult {
 
     private long row;
     private String message;
+    private boolean isOk;
 
     /**
      * @return The failed record's row
@@ -23,6 +24,13 @@ public class CsvValidationFailure {
         return message;
     }
 
+    /**
+     * @return true if the validation passed for this row
+     */
+    public boolean isOk() {
+        return isOk;
+    }
+
     public static Builder builder() {
         return new Builder();
     }
@@ -30,12 +38,14 @@ public class CsvValidationFailure {
     public static class Builder {
         private long row;
         private String message;
+        private boolean isOk;
 
-        public CsvValidationFailure build() {
-            final CsvValidationFailure failure = new CsvValidationFailure();
-            failure.row = row;
-            failure.message = message;
-            return failure;
+        public CsvValidationResult build() {
+            final CsvValidationResult result = new CsvValidationResult();
+            result.row = row;
+            result.message = message;
+            result.isOk = isOk;
+            return result;
         }
 
         public Builder row(final long row) {
@@ -45,6 +55,11 @@ public class CsvValidationFailure {
 
         public Builder message(final String message) {
             this.message = message;
+            return this;
+        }
+
+        public Builder isOk(final boolean isOk) {
+            this.isOk = isOk;
             return this;
         }
     }
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
@@ -1,8 +1,8 @@
 package org.opentestsystem.rdw.admin.model;
 
 /**
- * This model represents a CSV Validation failure containing the failing row
- * and failure message.
+ * This model represents a CSV Validation result containing the ending row
+ * and success or failure message
  */
 public class CsvValidationResult {
 
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
@@ -6,7 +6,7 @@ import org.apache.commons.csv.CSVRecord;
 import org.opentestsystem.rdw.admin.common.model.Organization;
 import org.opentestsystem.rdw.admin.common.security.PermissionScope;
 import org.opentestsystem.rdw.admin.model.CsvSubject;
-import org.opentestsystem.rdw.admin.model.CsvValidationFailure;
+import org.opentestsystem.rdw.admin.model.CsvValidationResult;
 import org.opentestsystem.rdw.admin.repository.SchoolRepository;
 import org.opentestsystem.rdw.admin.service.CsvValidationService;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -16,6 +16,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -44,9 +45,14 @@ public class DefaultCsvValidationService implements CsvValidationService {
             HeaderUser
     );
 
-    private static final String ValidationFormat = "Row: %1$d Failure: %2$s\n";
+    private static final String ValidationFailureFormat = "Row: %1$d Failure: %2$s\n";
+    private static final String ValidationSuccessFormat = "%1$d group%2$s accepted for %3$s (%4$s).";
+
     private static final String MoreFailuresFormat = "With %1$d additional validation failures\n";
-    private static final int MaxFailureMessages = 100;
+    private static final String MoreSuccessesFormat = "With %1$d additional schools\n";
+
+
+    private static final int MaxMessages = 100;
     static final int MaxStudentsPerGroup = 200;
 
     private final SchoolRepository schoolRepository;
@@ -57,31 +63,31 @@ public class DefaultCsvValidationService implements CsvValidationService {
     }
 
     @Override
-    public List<CsvValidationFailure> validateHeaders(final List<String> csvHeaders) {
-        final List<CsvValidationFailure> failures = new ArrayList<>();
+    public List<CsvValidationResult> validateHeaders(final List<String> csvHeaders) {
+        final List<CsvValidationResult> results = new ArrayList<>();
 
         for (int i = 0; i < RequiredHeaders.size(); i++) {
             if (!RequiredHeaders.get(i).equalsIgnoreCase(csvHeaders.get(i))) {
-                failures.add(CsvValidationFailure.builder()
+                results.add(CsvValidationResult.builder()
                         .row(0)
                         .message("Invalid headers. Headers must be in order: [" +
                                 Joiner.on(",").join(RequiredHeaders) +
                                 "]")
+                        .isOk(false)
                         .build());
                 break;
             }
         }
-        return failures;
+        return results;
     }
 
     @Override
-    public List<CsvValidationFailure> validateRecords(final Iterator<CSVRecord> recordIterator,
-                                                      final PermissionScope permissionScope) {
-        final List<CsvValidationFailure> validationFailures = new ArrayList<>();
+    public List<CsvValidationResult> validateRecords(final Iterator<CSVRecord> recordIterator,
+                                                     final PermissionScope permissionScope) {
+        final List<CsvValidationResult> validationResults = new ArrayList<>();
 
-        final Set<String> availableSchools = schoolRepository.findAll(permissionScope).stream()
-                .map(Organization::getNaturalId)
-                .collect(Collectors.toSet());
+        final List<Organization> availableSchools = schoolRepository.findAll(permissionScope).stream()
+                .collect(Collectors.toList());
 
         final Set<String> completedSchools = new HashSet<>();
         final Set<String> completedSchoolGroupYears = new HashSet<>();
@@ -90,7 +96,8 @@ public class DefaultCsvValidationService implements CsvValidationService {
         String currentGroup = null;
         String currentSchoolGroupYear = null;
         CsvSubject currentSchoolGroupYearSubject = null;
-        int studentCount = 0;
+        int studentCountPerGroup = 0;
+        int groupCountPerSchool = 0;
         long recordNumber = 0;
 
         while (recordIterator.hasNext()) {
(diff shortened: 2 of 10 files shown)


webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
```diff
@@ -4,10 +4,11 @@
  * This model represents a CSV Validation failure containing the failing row
  * and failure message.
```
bgsavage: This is why I don't like comments... I'll update....

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -121,68 +134,91 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
...
             if (isNotBlank(record.get(HeaderSSID))) {
-                studentCount++;
+                studentCountPerGroup++;
             }
         }
-        validateStudentCount(validationFailures, currentSchool, currentGroup, studentCount, recordNumber);
+        validateStudentCount(validationResults, currentSchool, currentGroup, studentCountPerGroup, recordNumber);
+
+        if(allOk(validationResults)) {
+            validationResults.add(newSuccessValidationResult(recordNumber, groupCountPerSchool, getSchoolOrganization(availableSchools, currentSchool)));
```
bgsavage: Only adding additional successes if there haven't been any failures to this point.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -101,17 +108,23 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
...
+            validateKeyValues(validationResults, school, group, schoolYear, recordNumber);
+            if (!allOk(validationResults)) continue;
 
             //Validate the School
             if (currentSchool == null || !currentSchool.equals(school)) {
-                if (currentSchool != null) completedSchools.add(currentSchool);
+                if (currentSchool != null) {
+                    completedSchools.add(currentSchool);
+                    if(allOk(validationResults)) {
+                        validationResults.add(newSuccessValidationResult(recordNumber, groupCountPerSchool, getSchoolOrganization(availableSchools, currentSchool)));
```
bgsavage, line 119: Only adding new successes if there are no failures currently.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -121,68 +134,91 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
...
-        validateStudentCount(validationFailures, currentSchool, currentGroup, studentCount, recordNumber);
+        validateStudentCount(validationResults, currentSchool, currentGroup, studentCountPerGroup, recordNumber);
+
+        if(allOk(validationResults)) {
+            validationResults.add(newSuccessValidationResult(recordNumber, groupCountPerSchool, getSchoolOrganization(availableSchools, currentSchool)));
+            return validationResults;
+        }
 
-        return validationFailures;
+        return  validationResults.stream().filter(x -> !x.isOk()).collect(Collectors.toList());
```
bgsavage: So this method will only return all failures or all successes.  It will never return a mix of successes / failures because we don't have any kind of partial success.  
> If we go down we go down together.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -195,50 +231,72 @@ private Integer getAsInt(final CSVRecord row, final String header) {
...
+                .isOk(true)
+                .build();
+    }
+
+    private static boolean allOk(final List<CsvValidationResult> validationResults) {
+        return validationResults.stream().allMatch(CsvValidationResult::isOk);
+    }
+
+    private static Organization getSchoolOrganization(final List<Organization> availableSchools, final String naturalId) {
+       return availableSchools.stream().filter(x -> Objects.equals(x.getNaturalId(), naturalId)).findFirst().get();
```
bgsavage, line 299: This `.get()` should be safe in that it will only get called in case all validations are successes, one of those validations being the school natural id.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
```diff
@@ -82,25 +83,31 @@ public StudentGroupBatch upload(final User user, final MultipartFile file) {
...
-                validationFailures.addAll(validate(digestStream, permissionScope));
+                validationResults.addAll(validate(digestStream, permissionScope));
             }
 
             final String digest = Hex.encodeHexString(messageDigest.digest()).toUpperCase();
-            if (validationFailures.isEmpty()) {
+            List<CsvValidationResult> failures = validationResults
+                    .stream()
+                    .filter(x -> !x.isOk())
+                    .collect(Collectors.toList());
```
bgsavage, line 97: I suppose technically I could just look at the first one as they should be the same, but just for thorough defensiveness in case something slips through in a later code update?

webapp/src/main/webapp/src/app/groups/groups.component.html
```diff
@@ -24,8 +24,10 @@
                     placeholder="{{'prompt.select' | translate}}"
                     filterPlaceholder="{{'labels.groups.search-school' | translate}}">
           <ng-template let-school pTemplate="item">
-            <div class="ui-helper-clearfix" style="position: relative;height: 25px;">
-              <div>{{school.name}} <span class="gray-dark font-weight-light">({{school.naturalId}})</span></div>
+            <div class="ui-helper-clearfix">
+              <div>{{school.name}}
+                <div class="font-weight-light small">{{school.naturalId}}</div>
+              </div>
```
bgsavage, line 30: unrelated UI update from previous PR on school natural id.

webapp/src/main/webapp/src/app/groups/import/import-table/import-table.component.html
```diff
@@ -3,29 +3,23 @@
...
   [paginator]="imports.length > 200"
   expandableRows="true"
+  rowExpandMode="single"
   [value]="imports"
   emptyMessage="{{'labels.groups.import.empty' | translate }}"
   sortField="id"
   sortOrder="-1"
   tableStyleClass="table table-striped table-hover"
   [tableStyle]="{'table-layout':'auto'}" >
+  <p-column expander="true" styleClass="col-icon"></p-column>
```
bgsavage, line 13: Expander column now that we always expand.

webapp/src/main/webapp/src/app/groups/import/import-table/import-table.component.html
```diff
@@ -3,29 +3,23 @@
...
   <p-column field="status" header="{{'labels.groups.import.cols.status' | translate }}" [sortable]="true">
     <ng-template let-import="rowData" pTemplate="body">
-      <span *ngIf="import.isOk" class="label border-only green">
+      <a href="javascript:void(0)" *ngIf="import.isOk" class="label border-only green" (click)="dt.toggleRow(import)" >
```
bgsavage, line 16: You can also click the status to expand / collapse the row.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
esotrope: ```toImmutableList()```?

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -57,31 +63,31 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
...
-                                                      final PermissionScope permissionScope) {
-        final List<CsvValidationFailure> validationFailures = new ArrayList<>();
+    public List<CsvValidationResult> validateRecords(final Iterator<CSVRecord> recordIterator,
+                                                     final PermissionScope permissionScope) {
+        final List<CsvValidationResult> validationResults = new ArrayList<>();
 
-        final Set<String> availableSchools = schoolRepository.findAll(permissionScope).stream()
-                .map(Organization::getNaturalId)
-                .collect(Collectors.toSet());
+        final List<Organization> availableSchools = schoolRepository.findAll(permissionScope).stream()
```
esotrope: I feel like guava would have a one-liner for this that isn't as resource wasteful as using a stream

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -121,68 +134,91 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
...
+        if (failures.size() > MaxMessages) {
+            messageBuilder.append(String.format(MoreFailuresFormat, failures.size() - MaxMessages));
+        }
+        return messageBuilder.toString();
+    }
+
+    @Override
+    public String toSuccessMessage(final List<CsvValidationResult> successes) {
+        final StringBuilder messageBuilder = new StringBuilder();
+        final List<CsvValidationResult> subList = successes.size() < MaxMessages ? successes : successes.subList(0, MaxMessages);
```
esotrope: nit: variable name should describe what is in the list

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -195,50 +231,72 @@ private Integer getAsInt(final CSVRecord row, final String header) {
...
                     .row(recordNumber)
                     .message("School [" + school + "] can not be found")
+                    .isOk(false)
                     .build());
         }
         if (completedSchools.contains(school)) {
-            validationFailures.add(CsvValidationFailure.builder()
+            validationResults.add(CsvValidationResult.builder()
                     .row(recordNumber)
                     .message("School [" + school + "] is out-of-order: sort CSV by school, group, year")
```
esotrope, line 279: Why aren't these run through a ```MessageSource```?

esotrope:
@aaronsleeper Do we want the chevron row expanders to be buttons in admin too? Wondering how that would work with this first column being a status label.

esotrope (approved):
Looks good to me. When possible we should add a message source and switch the collectors/list builders to immutables.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
esotrope: I'm with you. I like 'self documenting' code and comments when things are not clear because of this kind of maintenance difficulty as well.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
bgsavage: I actually don't need this stream at all anymore since we are no longer mapping just the `naturalId`

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -121,68 +134,91 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
...
-            messageBuilder.append(String.format(MoreFailuresFormat, failures.size() - MaxFailureMessages));
+        if (failures.size() > MaxMessages) {
+            messageBuilder.append(String.format(MoreFailuresFormat, failures.size() - MaxMessages));
+        }
+        return messageBuilder.toString();
+    }
+
+    @Override
+    public String toSuccessMessage(final List<CsvValidationResult> successes) {
+        final StringBuilder messageBuilder = new StringBuilder();
```
malaffoon: I think a StringJoiner would make this slightly more readable.

bgsavage: If java string builders had an `appendLine` it would also be slightly more readable :) 

I can use a StringJoiner instead though.

bgsavage:
I'm not sure if a message source is needed, but definitely can add immutability.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
```diff
@@ -1,13 +1,14 @@
...
- * and failure message.
+ * This model represents a CSV Validation result containing the ending row
+ * and success or failure message
  */
-public class CsvValidationFailure {
+public class CsvValidationResult {
 
     private long row;
     private String message;
+    private boolean isOk;
```
wtritch, line 11: Nit: Generally in java beans the property is named "ok" and the accessor is named isOk()

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -101,17 +108,23 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
             final CsvSubject subject = CsvSubject.valueOfCaseInsensitive(record.get(HeaderSubject));
             recordNumber = record.getRecordNumber();
 
-            validateKeyValues(validationFailures, school, group, schoolYear, recordNumber);
-            if (!validationFailures.isEmpty()) continue;
+            validateKeyValues(validationResults, school, group, schoolYear, recordNumber);
+            if (!allOk(validationResults)) continue;
```
wtritch, line 112: Hmmm, not a fan of repeatedly iterating over the list...

wtritch: I think I'd rather see a container that has the successes and failures tracked separately.  Thoughts?

bgsavage: That's a great idea since I'm already returning separate collections anyways.

PR #47, opened 2017-09-13
Now using separate collections to track failures and successes when v…

bgsavage:
…alidation the group upload.

Thanks for the suggestion @wtritch


wtritch (approved):
LGTM =)

PR #48, opened 2017-09-13
property placeholders in datasource url

malaffoon:


webapp/src/main/resources/application.yml
```diff
@@ -30,13 +30,19 @@ server:
 
 spring:
   datasource:
-    url: jdbc:mysql://localhost:3306/warehouse?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8
+    url: "\
+      jdbc:mysql://${spring.datasource.url-server:localhost:3306}/${spring.datasource.url-schema:reporting}\
```
malaffoon: oops ... will fix

esotrope (approved):
"typo" at least means you attempted to type something and failed. but you didn't even type it to begin with : P

bgsavage (approved):

webapp/src/main/resources/application.yml
```diff
@@ -30,13 +30,19 @@ server:
 
 spring:
   datasource:
-    url: jdbc:mysql://localhost:3306/warehouse?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8
+    url: "\
+      jdbc:mysql://${spring.datasource.url-server:localhost:3306}/${spring.datasource.url-schema:warehouse}\
+      ?useSSL=${spring.datasource.use-ssl:false}\
+      &useLegacyDatetimeCode=${spring.datasource.use-legacy-datetime-code:false}\
+      &characterEncoding=${spring.datasource.character-encoding:utf8}\
```
bgsavage, line 37: Where do these get set?

malaffoon: Mostly they never get set since the defaults are good. But the individual params can be set in the configuration yml. An example is the test profile below ...

webapp/src/test/resources/application-test.yml
```diff
@@ -1,14 +1,10 @@
 spring:
   datasource:
-    url: jdbc:mysql://localhost:3306/warehouse_test?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8
+    url-schema: warehouse_test
```
malaffoon, line 3: ... instead of copying & pasting and then maintaining the full connection string, this lets just the one variable be set.

PR #49, opened 2017-09-20
Defect/incorrect max group validation

esotrope:
1. Corrected max group validation check
2. Style checked code (made variable naming more consistent, ran formatter)
3. Added ```CsvValidationResult``` factory methods ```success``` and ```failure``` to reduce duplicate code and make method bodies more succinct.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
@@ -1,5 +1,7 @@
 package org.opentestsystem.rdw.admin.model;
 
+import org.opentestsystem.rdw.admin.common.model.Organization;
+
 /**
  * This model represents a CSV Validation result containing the ending row
  * and success or failure message
@@ -8,7 +10,7 @@ public class CsvValidationResult {
 
     private long row;
     private String message;
-    private boolean isOk;
+    private boolean ok;
 
     /**
      * @return The failed record's row
@@ -28,23 +30,40 @@ public class CsvValidationResult {
      * @return true if the validation passed for this row
      */
     public boolean isOk() {
-        return isOk;
+        return ok;
     }
 
     public static Builder builder() {
         return new Builder();
     }
 
+    public static CsvValidationResult success(final long row, final String message) {
+        return CsvValidationResult.builder()
+                .row(row)
+                .message(message)
+                .ok(true)
+                .build();
+    }
+
+    public static CsvValidationResult failure(final long row, final String message) {
+        return CsvValidationResult.builder()
+                .row(row)
+                .message(message)
+                .ok(false)
+                .build();
+    }
+
     public static class Builder {
+
         private long row;
         private String message;
-        private boolean isOk;
+        private boolean ok;
 
         public CsvValidationResult build() {
             final CsvValidationResult result = new CsvValidationResult();
             result.row = row;
             result.message = message;
-            result.isOk = isOk;
+            result.ok = ok;
             return result;
         }
 
@@ -58,8 +77,8 @@ public class CsvValidationResult {
             return this;
         }
 
-        public Builder isOk(final boolean isOk) {
-            this.isOk = isOk;
+        public Builder ok(final boolean ok) {
+            this.ok = ok;
             return this;
         }
     }
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
@@ -1,7 +1,5 @@
 package org.opentestsystem.rdw.admin.model;
 
-import org.opentestsystem.rdw.admin.common.model.Organization;
-
 /**
  * This model represents a CSV Validation result containing the ending row
  * and success or failure message
@@ -37,6 +35,13 @@ public class CsvValidationResult {
         return new Builder();
     }
 
+    /**
+     * Creates a successful {@link CsvValidationResult} for the given row and message
+     *
+     * @param row the row number that was validated
+     * @param message the success message
+     * @return a {@link CsvValidationResult} for the given row and message and sets {@link CsvValidationResult#isOk()} to <code>true</code>
+     */
     public static CsvValidationResult success(final long row, final String message) {
         return CsvValidationResult.builder()
                 .row(row)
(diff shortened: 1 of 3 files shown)


webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -116,9 +105,7 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
             if (currentSchool == null || !currentSchool.equals(school)) {
                 if (currentSchool != null) {
                     completedSchools.add(currentSchool);
-                    if(failures.isEmpty()) {
```
esotrope, line 119: This if condition can never be false because of the above ```if (!failures.isEmpty()) continue;``` statement

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
```diff
@@ -161,40 +148,35 @@ public DefaultCsvValidationService(final SchoolRepository schoolRepository) {
...
                                       final int studentCount,
                                       final long recordNumber) {
-        if (studentCount < MaxStudentsPerGroup) return;
-
-        validationFailures.add(CsvValidationResult.builder()
-                .row(recordNumber)
-                .message("School [" + school + "] Group [" + group + "] contains " + studentCount + " students.  Maximum students allowed per group is " + MaxStudentsPerGroup)
-                .isOk(false)
-                .build());
+        if (studentCount > MaxStudentsPerGroup) {
```
esotrope, line 167: The actual bug fix

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
```diff
@@ -175,35 +176,25 @@ private void archiveFile(final MultipartFile file,
...
-            validationFailures.clear();
-            validationFailures.add(CsvValidationResult.builder()
-                    .row(0)
-                    .message("File does not appear to be a valid CSV with a header row")
-                    .build());
+
+            failures.addAll(validationService.validateHeaders(orderedHeaders));
+            failures.addAll(validationService.validateRecords(recordIterator, permissionScope));
+
+        } catch (final IOException exception) {
```
esotrope, line 189: What actually throws the IOException? recordIterator.next()? My guess is the input stream open/interruption can trigger it.
Does it make sense to handle this like the IllegalArgumentException and return a collection of messages just containing this message?
Alternatively, if this is an exception that can be thrown on any row shouldn't we pass on the row number?

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
malaffoon: I like most of the other changes you made ... made things more readable. But why change away from guard-code style here? Just to clarify, if the code remained guard-code style, the fix would be `<` -> `<=`?

malaffoon (approved):

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
bgsavage: Yeah, I question `!failures.isEmpty() continue` as it doesn't report all failures in that case.  I thought the original intent was for it to report all failures from Will's demo.  I think removing this defensive check can cause issues later should we decide again to report all failures.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
bgsavage: I'm good with removing the guard-code style here in favor of consistency with the other validations.

webapp/src/main/java/org/opentestsystem/rdw/admin/model/CsvValidationResult.java
```diff
@@ -28,23 +28,54 @@ public String getMessage() {
...
+     * @param row the row number that was validated
+     * @param message the success message
+     * @return a {@link CsvValidationResult} for the given row and message and sets {@link CsvValidationResult#isOk()} to <code>false</code>
+     */
+    public static CsvValidationResult failure(final long row, final String message) {
+        return CsvValidationResult.builder()
+                .row(row)
+                .message(message)
+                .ok(false)
+                .build();
```
bgsavage, line 65: Nice 👍

bgsavage (approved):

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
esotrope: @malaffoon Yes ```<=``` would have fixed the bug.
Also, I changed away from the guard to be consistent with the rest of the methods like @bgsavage said.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java
esotrope: Well, it doesn't belong in the current implementation. I can add a comment though in place of the duplicate if check

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
esotrope: @wtritch any insight on this?

wtritch: I know this exception can be thrown if there is a problem opening/reading the uploaded file.  I would assume that could happen anywhere during CSV streaming, although it's pretty unlikely after the initial load.  It's definitely unrecoverable, so I think returning at this point is ok.
I'm wondering if we actually want to change the IllegalArgumentException handling to also append to failures, rather than assume it's the one-and-only.  I'd hate to throw any other known failures on the floor when we could instead send them back to the user.

wtritch (approved):

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
esotrope: sounds good. I'll make them both append and return the current state.

PR #50, opened 2017-09-20
Setting a height instead of using padding alone allows drag and drop …

bgsavage:
…to work in IE11.

This is the height of the box after padding is applied.

### DIFF
--- a/webapp/src/main/webapp/src/styles.less
+++ b/webapp/src/main/webapp/src/styles.less
@@ -139,6 +139,7 @@ footer {
 .my-drop-zone {
   border: dotted 3px lightgray;
   border-radius: 5px;
+  height: 118px; /* Required for drag & drop to work in IE11 */
 }
 
 .nv-file-over {


webapp/src/main/webapp/src/styles.less
```diff
@@ -139,6 +139,7 @@ footer {
 .my-drop-zone {
   border: dotted 3px lightgray;
   border-radius: 5px;
+  height: 118px; /* Required for drag & drop to work in IE11 */
```
bgsavage, line 142: After the scrum call I tried hitting the ng2-file-upload site to see if the demo page's drag and drop was working in IE11, and to my astonishment, it was.  So I started investigating what was different between their implementation and mine, and found they were using a`.well` which applies a min-height to the element.

esotrope, line 142: Would it still work with min-height?
I'm thinking if there is another more verbose message (like if we later switch to Vietnamese) occupying the same space we wouldn't want overflow

bgsavage: There's a lot of room to grow...

image

The drag and drop area might be restricted to the min-height area in IE11, would need to test.

esotrope (approved):

PR #51, opened 2017-09-20
Updated SamlUserDetailsService to no longer throw an exception when n…

bgsavage:
…o permissions are found, and instead to return a User with no permissions.  This allows the app to continue to load the user into the SecurityContextHolder so that logout can be made possible through the angular app.

Updated access-denied to page to current implementation in Reporting.

### DIFF
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/security/saml/SamlUserDetailsService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/security/saml/SamlUserDetailsService.java
@@ -43,8 +43,7 @@ public class SamlUserDetailsService implements SAMLUserDetailsService {
 
         // Secures the application by returning a user without any permissions or authorities.
         if (grants.isEmpty()) {
-            final String message = String.format("Grants do not permit access to tenant for user \"%s\"", username);
-            logger.warn(message);
+            logger.warn(String.format("Grants do not permit access to tenant for user \"%s\"", username));
 
             return User.builder()
                     .id(id)
--- a/webapp/src/main/java/org/opentestsystem/rdw/admin/security/saml/SamlUserDetailsService.java
+++ b/webapp/src/main/java/org/opentestsystem/rdw/admin/security/saml/SamlUserDetailsService.java
@@ -5,6 +5,8 @@ import org.opentestsystem.rdw.admin.security.Permission;
 import org.opentestsystem.rdw.admin.security.User;
 import org.opentestsystem.rdw.admin.security.auth.AuthorizationService;
 import org.opentestsystem.rdw.security.Grant;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
@@ -21,6 +23,7 @@ public class SamlUserDetailsService implements SAMLUserDetailsService {
 
     private static final String REDACTED = "[REDACTED]";
     private AuthorizationService authorizationService;
+    private static final Logger logger = LoggerFactory.getLogger(SamlUserDetailsService.class);
 
     @Autowired
     public SamlUserDetailsService(@NotNull AuthorizationService authorizationService) {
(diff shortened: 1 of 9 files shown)


webapp/src/main/webapp/src/app/app.component.html
```diff
@@ -46,12 +46,6 @@
 <div class="page-view pt-md">
   <div class="container">
     <router-outlet></router-outlet>
-
-    <div *ngIf="userHasAccess === false">
-      <div class="alert alert-danger">
-        {{ 'messages.no-access' | translate }}
-      </div>
-    </div>
```
bgsavage, line 54: We were just showing an alert before.  No longer needed.

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -21,26 +22,30 @@ export class AppComponent {
...
   ngOnInit() {
     this._userService.getCurrentUser().subscribe(user => {
 
       this._user = user;
+      let userHasAccess = user.permissions.some(permission => RequiredPermissions.indexOf(permission) !== -1);
 
-      this.userHasAccess = user.permissions.some(permission => permission == "GROUP_WRITE");
+      if (!userHasAccess) {
+        this.router.navigate([ 'access-denied' ]);
+      }
```
bgsavage, line 45: Now here in the AppComponent ngInit if we find the user has none of the RequiredPermissions then we forward them to access-denied.

malaffoon (approved):
I'm assuming this is copied over from reporting so will work swimmingly first go-around.

webapp/src/main/webapp/src/app/app.component.ts
esotrope: I feel like there would be a more angular way to do this than in ngOnInit. Can the "canActivate" route hooks not redirect you based on absence of permissions?

webapp/src/main/webapp/src/app/app.component.ts
```diff
@@ -21,26 +22,30 @@ export class AppComponent {
...
-    translate.use(languages.indexOf(translate.getBrowserLang()) != -1 ? translate.getBrowserLang() : defaultLanguage);
+    translate.use(SupportedLanguages.indexOf(translate.getBrowserLang()) != -1 ? translate.getBrowserLang() : defaultLanguage);
 
   }
 
   ngOnInit() {
     this._userService.getCurrentUser().subscribe(user => {
 
       this._user = user;
+      let userHasAccess = user.permissions.some(permission => RequiredPermissions.indexOf(permission) !== -1);
```
esotrope: The name required permissions make me think it would have an AND relationship instead of the OR impl here.
Might make sense to make a utility similar to spring's hasAnyAuthority() method so we can make quick sense of what is going on here.
```has(user.permissions, "GROUP_WRITE")``` or ```hasAnyOf(user.permissions, ["GROUP_WRITE"])```

File: webapp/src/main/java/org/opentestsystem/rdw/admin/security/saml/SamlUserDetailsService.java [hunk 4]
```diff
@@ -38,10 +41,19 @@ public Object loadUserBySAML(final SAMLCredential credential) throws UsernameNot
 
         final Set<Grant> grants = authorizationService.getGrants(encodedGrants);
 
-		// Secures the application by throwing exception if the user has no authority to access data for this state
+        // Secures the application by returning a user without any permissions or authorities.
         if (grants.isEmpty()) {
-            throw new RuntimeException(String.format(
-                    "Grants do not permit access to tenant for user \"%s\"", username));
+            final String message = String.format("Grants do not permit access to tenant for user \"%s\"", username);
+            logger.warn(message);
```
esotrope: could be simplified to:
```logger.warn("Grants do not permit access to tenant for user \"{}\"", username);```

esotrope (approved):

webapp/src/main/webapp/src/app/app.component.ts
bgsavage: I did something like that more along those lines in Reporting since we had a home component and a home route.  Here, we don't really have that and start navigation on a functional page.  Until we have some root home route I don't think it makes sense to do that now.

And as for doing it on AppComponent, you can't because it's the root and is not bound to a ui route.

esotrope: Hm. Wonder if there was a better pattern we could have followed where the app component just acted as an empty shell and the whole app view is delivered through the routes.

webapp/src/main/webapp/src/app/app.component.ts
bgsavage: I can rename it just to `AdminPermissions` then.  My intent is to self document in one spot what our supported languages are and what are the permissions will be needed.

esotrope: K. We may want to introduce an ```application-settings.ts``` similar to our spring ```applicaiton.yml```
Thoughts?

PR #52, opened 2017-09-25
DWR-845, changes for district/school groups

malaffoon:
Initial changes to admin for district/school groups. Most of the change is from moving code into common (to share with reporting). I'll try to call out the interesting bits.

I will update this to bump common dependency when that review is blessed.

NOTE: there are more code changes before all this is working. All the queries that use Permission need to be changed to use the extended PermissionScope.


webapp/src/main/java/org/opentestsystem/rdw/admin/Application.java
```diff
@@ -24,7 +25,8 @@
         EndpointAutoConfiguration.class,
         StatusConfiguration.class,
         YamlPropertiesConfigurator.class,
-        DataSourceConfiguration.class
+        DataSourceConfiguration.class,
+        JdbcOrganizationRepository.class
```
malaffoon, line 29: This is a @Component in RDW_Common. This is the least impactful way to force it into this project; component scanning caused some odd behavior with the controllers.

webapp/src/main/java/org/opentestsystem/rdw/admin/security/auth/DefaultAuthorizationService.java
```diff
@@ -3,13 +3,13 @@
 import com.google.common.collect.ImmutableSet;
```
malaffoon, line 3: This class is where most of the real changes are. 

Yeah, i could try to pull this whole class into RDW_Common but it started to get messy. Maybe next go-around.

webapp/src/main/resources/application.sql.yml
```diff
@@ -3,12 +3,24 @@ sql:
     findWarehouseImports: >-
```
malaffoon: oops, these shouldn't be here any more

webapp/src/test/java/org/opentestsystem/rdw/admin/security/auth/DefaultAuthorizationServiceTest.java
```diff
@@ -77,36 +73,37 @@ public void getGrantsShouldOnlyIncludeGrantsForTheSpecifiedState() throws Except
...
 
     @Test
     public void getPermissionsShouldCorrectlyDerivePermissionsFromGrants() throws Exception {
         final List<Grant> grants = ImmutableList.of(
-                Grant.builder().entityId("a").role("a").entityLevel(STATE).stateId("a").build(),
-                Grant.builder().entityId("a").role("b").entityLevel(DISTRICT).stateId("a").districtId("a").build(),
-                Grant.builder().entityId("a").role("c").entityLevel(INSTITUTION).stateId("a").districtId("a").institutionId("a").build(),
-                Grant.builder().entityId("a").role("d").entityLevel(DISTRICT).stateId("a").districtId("a").build(),
-                Grant.builder().entityId("a").role("d").entityLevel(INSTITUTION).stateId("a").districtId("a").institutionId("a").build()
+                Grant.fromString("|a|a|STATE|||||a||"),
```
malaffoon, line 83: i didn't retain the Grant builder so i had to convert these to tenancy chains

common/src/test/java/org/opentestsystem/rdw/admin/common/test/support/PermissionScopes.java
```diff
@@ -16,11 +16,11 @@ public static PermissionScope statewide() {
     }
 
     public static PermissionScope districts(final Long... ids) {
-        return new PermissionScope(Arrays.asList(ids), null);
+        return new PermissionScope(Arrays.asList(ids), null, null, null);
```
esotrope, line 19: uh oh i think we broke our self-imposed 3 param limit and need a builder now : p

esotrope (approved):

common/src/test/java/org/opentestsystem/rdw/admin/common/test/support/PermissionScopes.java
malaffoon: Dang, i was hoping nobody would catch that.

esotrope: Hard to miss with all those consecutive nulls. I thought you were instantiating an array x J

jtreuting: I believe I said 5 to start and someone (.... @malaffoon ) thought 3 was better

webapp/src/main/resources/application.sql.yml
```diff
@@ -3,12 +3,6 @@ sql:
...
       SELECT i.id, i.digest FROM import i limit 10;
 
-  district:
-    findAllIds: >-
```
jtreuting, line 7: where did the findAllIds go?  not used or moved to the common application.sql.yml?

malaffoon: moved into common

PR #54, opened 2017-10-03
Add Spring Session support for distributed session storage

wtritch:
This PR handles DWR-1025 for adding spring-session support to RDW_Admin.  If there are any "gotchas" from the RDW_Reporting implementation that I've missed, please let me know.


common/src/main/java/org/opentestsystem/rdw/admin/common/security/PermissionScope.java
```diff
@@ -2,13 +2,14 @@
...
 
+import java.io.Serializable;
 import java.util.Collection;
 import java.util.Set;
 
 /**
  * Used to limit database query results
  */
-public class PermissionScope {
+public class PermissionScope implements Serializable {
```
wtritch, line 12: The Serializable interfaces caught me off-guard.  I didn't realize they were required until I ran the app against a local redis service.

webapp/src/main/java/org/opentestsystem/rdw/admin/security/session/SecuredRedisConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.session.data.redis.config.ConfigureRedisAction;
+
+/**
+ * This class allows spring-redis to work in a secured redis environment
+ */
+@Configuration
+@ConditionalOnProperty(name = "spring.session.store-type", havingValue = "redis")
+public class SecuredRedisConfiguration {
```
wtritch, line 13: Taken from RDW_Reporting.

webapp/src/main/resources/application.yml
```diff
@@ -62,6 +65,9 @@ spring:
           destination: GROUPS
           group: default
 
+  session:
+    store-type: none
```
wtritch, line 69: By default, when running locally, we won't use distributed sessions.

common/src/main/java/org/opentestsystem/rdw/admin/common/security/PermissionScope.java
esotrope: These are required now with the new redis serialization?

esotrope (approved):

common/src/main/java/org/opentestsystem/rdw/admin/common/security/PermissionScope.java
wtritch: Yeah, because redis by default uses a JdkSerializationRedisSerializer to serialize session information.  That serializer then uses an ObjectOutputStream/ObjectInputStream to serialize/deserialize session information using Java serialization.  So only Serializable thingies are allowed to exist in the session store.

wtritch: Changing it to use a Jackson/JSON serializer would require some custom serializers/deserializers to handle the SAML XML payloads.

esotrope: Ok I remember that now. I thought we changed it to use JSON for some reason but i remember now that I couldn't get that working.

malaffoon, line 12: Ugh, merging into develop will be annoying since these have been refactored to be shared.

PR #56, opened 2017-10-04
Task/hotfix 1.0.1 merge

wtritch:
This PR merges hotfix 1.0.1 back into "develop."


build.gradle
```diff
@@ -37,7 +37,7 @@ String dockerPrefix = "smarterbalanced"
 
 allprojects {
     group = 'org.opentestsystem.rdw.reporting'
-    version = '1.0.0-SNAPSHOT'
+    version = '1.0.1-SNAPSHOT'
```
malaffoon, line 40: personally i'd change this to 1.1.0 but i realize you are just merging changes

malaffoon:
Are you planning to move the shared code into RDW_Common security at some point? Or should we just create a JIRA for it?

build.gradle
wtritch: I meant to bring that up, our 'develop' versions seem to be a little behind.  I don't think this is the PR to fix it/them, but something to be aware of.

malaffoon (approved):

wtritch:
I don't want to merge it into RDW_Common, because that will add a dependency on SAML to RDW_Common.  I'd rather merge it into our upcoming RDW_Reporting/web-base or whatever module that will be shared by admin and reporting webapps

PR #63, opened 2017-10-10
DWR-1026 Handle an optional BOM character in uploaded CSVs

wtritch:
When uploading a CSV from IE11, a BOM character is added to the uploaded CSV.  This leads to a CSV header column of "u'\ufeff'group_name" rather than "group_name" which causes validation failures.

This PR wraps the InputStream in an apache commons-io BOMInputStream that strips BOM bytes from the stream if they exist.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,7 +76,7 @@ subprojects {
 
             // commonly used libraries
             dependency 'com.google.guava:guava:21.0'
-            dependency 'org.apache.commons:commons-io:1.3.2'
+            dependency 'commons-io:commons-io:2.5'
             dependency 'org.slf4j:slf4j-api:1.7.25'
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -76,7 +76,7 @@ subprojects {
 
             // commonly used libraries
             dependency 'com.google.guava:guava:21.0'
-            dependency 'org.apache.commons:commons-io:1.3.2'
+            dependency 'commons-io:commons-io:2.5'
```
wtritch, line 79: Bump to the latest apache commons-io version

malaffoon (approved):
Can we have a test that demonstrates the handling of a file with a BOM and non-ascii characters?

wtritch:
Good call, added a BOM validation test and appended Chinese characters to the test CSV values.

PR #65, opened 2017-12-05
detect and note newline sequence

malaffoon:
This is the hotfix for the problem with newline sequences in student group upload files. There is a matching patch to RDW_Ingest.

The approach is to detect the newline sequence during validation and store it with the file properties. Then the group processor can use that to set the expected newline during import.


webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
```diff
@@ -82,22 +85,26 @@ public StudentGroupBatch upload(final User user, final MultipartFile file) {
...
-        final List<CsvValidationFailure> validationFailures = new ArrayList<>();
 
         try {
+            final List<CsvValidationFailure> validationFailures = newArrayList();
+            final Map<String, String> fileProperties = newHashMap();
+            fileProperties.put("username", user.getUsername());
+
             try (final DigestInputStream digestStream = new DigestInputStream(file.getInputStream(), messageDigest)) {
-                validationFailures.addAll(validate(digestStream, permissionScope));
+                validate(digestStream, permissionScope, validationFailures, fileProperties);
```
malaffoon, line 95: Modified the signature of the helper to effectively return failures and properties.

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
```diff
@@ -82,22 +85,26 @@ public StudentGroupBatch upload(final User user, final MultipartFile file) {
...
                 logger.debug("Validated request: {}", file.getOriginalFilename());
-                archiveFile(file, digest, user);
+                archiveFile(file, digest, fileProperties);
                 studentGroupBatchBuilder
                         .digest(digest)
                         .status(ImportStatus.ACCEPTED);
             } else {
-                logger.debug("Request failed validation: {}", file.getOriginalFilename());
+                logger.info("Request failed validation: {}", file.getOriginalFilename());
+                archiveFile(file, digest, fileProperties);
```
malaffoon, line 107: sneaking in another change: archive the file even if the data is bad

jtreuting (approved):

webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultStudentGroupBatchService.java
agorina: Nice!

agorina (approved):

wtritch (approved):
LGTM

